PR #26, opened 2020-10-22
buildable with nimble

jiro4989:
Nim usually builds packages with [nimble](https://github.com/nim-lang/nimble).
This PR makes arturo buildable with nimble.

```bash
$ nimble build
$ nimble build -d:DEBUG
$ nimble build -d:VERBOSE
$ nimble install
```


drkameleon:
Hi!

And thanks a lot for the PR. I'm looking into it.
I just want to make sure everything works right first, mainly the Travis build that seems broken.

Once, it's all sorted out, I'll go through the change.

Regarding the use of `nimble`, I'm not against it at all - thanks a lot for the work again. I would just prefer to have the two build scripts side by side. The less dependencies on 3rd-party scripts, the more flexibility in the future I think...

drkameleon (approved):
I will approve it as-is and fix a couple of things later.

Thanks! :)

PR #31, opened 2020-10-29
fix: deprecated code (set-env)

jiro4989:
`set-env` is deprecated commands.
see: [GitHub Actions: Deprecating set-env and add-path commands](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/)

Using [Setting an environment variable](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable).


drkameleon (approved):
Perfect! :)

PR #256, opened 2021-03-23
Better handling of commandline args

drkameleon:
# Description

Currently command-line arguments are accessible via the `arg` constant. However, switch-type arguments are not handled correctly (and are instead processed by the interpreter itself - as if they were arguments to Arturo instead of the script in question).

Also, apart from the `arg` variable, we should be able to process more complex command-line arguments automatically. (Add a different global variable?)

## Type of change

- [X] Code cleanup
- [X] Bug fix (non-breaking change which fixes an issue)


drkameleon:
Ready to merge! 🚀

PR #318, opened 2021-05-01
Add support for complex value type + misc Numbers library fixes/additions

drkameleon:
# Description

A new `:complex` value type is to be added.

**A few premises:**

- There is no reason of a new direct literal constructor, e.g. `a+bi`
- New values will be constructed via the generic `to` converter, e.g. `to :complex [1.0 2.0] -> complex(1.0, 2.0) -> 1.0 + 2.0i`
- The new value should overload both the printable (`$`) and dump (for `inspect`) methods - preferably being printed like: `a + bi` (or `a+bi`?)
- We have to support at least the basic arithmetic operations: `+`, `-`, `*`, `/` (*with* their in-place counterparts)
- The only comparison operator to be considered is `=`
- We could also add support for this new type in a few relevant functions, already in the built-in library, e.g: `abs`, `log`, `ln`, along with the trigonometric ones

## Type of change

Please select the option(s) that is more relevant.


progamateur:
Another relevant function for :complex : angle
c: 1 + 1j 
angle c ; ->  0.7854 (45°)
angle function  would be equivalent to the atan2 function  ( as in matlab)
Btw : atan2 function  would be useful.

drkameleon:
> Another relevant function for :complex : angle
> c: 1 + 1j
> angle c ; -> 0.7854 (45°)
> angle function would be equivalent to the atan2 function ( as in matlab)
> Btw : atan2 function would be useful.

Numbers\atan2 has just been implemented! :) (And yes, it works even with integers, as pretty much every trigonometric-related function)

Screenshot 2021-05-01 at 13 04 42

Re: `angle`, since the name sounds a bit too common to waste it on something not-so-common, would `phase` do? (I've had a look around and noticed that it's quite common to use this for that particular functionality)

progamateur:
"phase" is ok, but as far as I know, "angle" is more used (matlab, numpy, ...)
Btw:  parameters order of atan2 ?  atan2 x y or atan2 y x

drkameleon:
> "phase" is ok, but as far as I know, "angle" is more used (matlab, numpy, ...)
> Btw: parameters order of atan2 ? atan2 x y or atan2 y x

ok, let's stick to "angle" then. And if there is any weird name collision in the future (we want this particular word to mean something else, we'll see then...) ;)

Re atan2: funny that you mention it, because I've *just* changed it to `atan2 y x`, right now! lol (it's more normal, isn't it?)

progamateur:
Yes this "order" is more common.
Btw: what about floating in single precision (32 bits) ? (Sorry it's another subject )

drkameleon:
> Yes this "order" is more common.
> Btw: what about floating in single precision (32 bits) ? (Sorry it's another subject )

Right now, the inner structure for `:integer`s is a 64-bit int (expandable automatically to GMP-powered bignums), and for `:floating`s, it's also 64-bit floats.

I have thought of extending the type system (mostly as part of a - very ambitious - project of going full self-hosted, and making Arturo compile-able), but I think it hasn't been one of my priorities.

drkameleon:
Ready to merge 🚀

PR #358, opened 2021-05-17
Add extensive :path support

drkameleon:
# Description

Right now, `:path`s and `:pathLabel`s support only constant indexes/keys.

We should be able to support any typed of indexes/keys, including whole expressions, like:

```red
arr: ["zero" "one" "two" "three"]
x: 1
print arr\[x] ; should print "one"
print arr\[x+1] ; this should print "two"
arr\[x+1]: "dos" ; this would set element x+1=2 of the array to "dos"
```

## Type of change

- [X] Code cleanup


drkameleon:
Ready to merge 🚀

PR #363, opened 2021-05-24
Add line number tracking for better errors

drkameleon:
# Description

Runtime Errors currently don't contain any source line information making it difficult to spot the errors. We should add this information.

The main idea is (based on @mikeparr's idea: https://discord.com/channels/765519132186640445/765519132186640448/846141805216923668):

- Have a new `Newline` value type
- When the parser encounter a newline (`\n`, `\r`, etc), store that as a `Newline` value, along with the necessary information (which line in the source we're currently at) + the source file path (in the future)
- When the evaluator encounters a `Newline` value, it will output an `EOL` bytecode
- When the exec function encounters an `EOL` the result will be to ignore it and internally update a `CurrentLine` global variable, which can then be used for error reporting (in case an error occurs, that is)

* Preferably, the whole thing should be easily excludable (e.g. have at least a BUILD flag that can disable this whole setup, in case we don't want to include it in production builds) and/or a command-line flag, that enables/disables it at the expense of some performance loss.

## Type of change

Please select the option(s) that is more relevant.

- [X] Code cleanup


drkameleon:
Ready to merge 🚀

PR #369, opened 2021-05-29
Support conversion from integer to any char

feihong:
# Description

This would make `to :char 20320` return `你` instead of junk char.

## Type of change

- [ ] Code cleanup


drkameleon:
First of all, welcome! 😃 

And... awesome job! I had missed this. (I guess I may have missed a couple of other things too related to Unicode, so... if you spot anything else, feel free to let me know or just make another PR.)

Since I myself work with different - non-English - charsets a lot (including 汉字), complete Unicode support has been a must since the very beginning. 

I'm merging it right now! Thanks! 😉 

P.S Don't bother at all about the 2 supposedly failing checks - they are of no importance, it seems like an issue with the checks running on a different fork. I'll have to look into it...

PR #375, opened 2021-06-03
Integrate three-value logic values

drkameleon:
# Description

Right now, `:boolean` values can be either `true` or `false`.

The aforementioned type will be replaced by a `:logical` type, which may also have support for a third value `maybe`.

All built-in functions supporting booleans should carefully migrated to the new model.

## Type of change

- [X] Code cleanup
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)


drkameleon:
Ready to merge 🚀

PR #379, opened 2021-06-04
Improved type safety

drkameleon:
# Description

Right now, type safety in Arturo is a weak point. We should begin by adding some constraints, which will lead to fewer bugs - but in a totally optional way.

For example, if we have a function that takes 2 arguments, we could implement the relevant syntax to *optionally* set the expected type for any of the given arguments.

**Example:**

```red
f: function [x :integer, y :floating][
    print x + y
]
f 2 10.0
```

**Will be equivalent to:**

```red
f: function [x, y][
    ensure -> is? :integer x 
    ensure -> is? :floating y
    print x + y
]
f 2 10.0
```

## Type of change

- [X] Code cleanup


drkameleon:
Ready to merge 🚀

Issue #207, opened 2021-03-08
REPL terminating when closing Webview window


drkameleon:
Originally referenced by @trekLOF: https://discord.com/channels/765519132186640445/765519132186640448/818415774876041246

----

**Describe the bug**
When using the REPL and opening a webview window (with the `webview` command), once the webview is terminated, the REPL quits as well.

**To Reproduce**

1. open the Arturo REPL: `$> arturo`
2. open a webview: `$> webview "hello!"`
3. close the webview

**Expected behavior**
We should return to the REPL and continue as usual.

**Additional context**
Opening a webview [triggers an app exit](https://github.com/arturo-lang/arturo/blob/master/src/library/Ui.nim#L120). This shouldn't happen - either at all, or when in REPL mode (how to detect that?).

PR #409, opened 2022-01-29
WIP: Better Webview & UI support

drkameleon:
# Description

Currently, we have [**webview-based UI**](https://arturo-lang.io/documentation/library/ui/webview/) support (based on an older version of [Serge Zaitsev's Webview](https://github.com/webview/webview) library, Chrome-based UI (basically, using the already-installed Chrome-compatible chrome to show any URL/page) & [**server**](https://arturo-lang.io/documentation/library/net/serve/) support (based roughly on Nim's standard library [AsyncHttpServer](https://nim-lang.org/docs/asynchttpserver.html)).

Both parts *do* work, with a few asterisks:

- we are currently using an older C-only version of the Webview library - which is a) good, because it's pure C, b) bad, because there are changes that are great and *should* be incorporated
- the distinction between the different library functions (`webview`, `serve`, `serve.chrome`) is a bit weird and confusing
- the integration between the "backend" (Arturo itself) and the webview is a bit messy (basically, it kinda works, but even I don't remember exactly how... lol) - this should be cleaned up, or re-written from scratch
- the server part and the webview don't work together - mainly, having to do with blocking threads (for example, for Grafito, the ideal scenario would be: start a server, and open up a webview - non-Chrome - window pointing to that server. No big deal. But it still doesn't work)

Given that the UI part (and especially webview-based UI functionality) seems to be *the* future of UI apps - or, at least, a good part of them, this has to be worked on.

Related: https://github.com/arturo-lang/arturo/pull/275
Fixes: #207 #261

## Type of change

- [X] Code cleanup
- [X] Bug fix (non-breaking change which fixes an issue)
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)


drkameleon:
A minimal server-webview coordination example that seems to be working fine (at least with the existing, older version of the Webview library) - for future reference:

```nim
import std/asynchttpserver
import std/asyncdispatch, std/threadpool

import src/helpers/webview

echo "current thread: " & $(getThreadId())

proc initWebview() =
    echo "webview thread: " & $(getThreadId())
    let wv = createWebview(
        title       = "My webview", 
        url         = "http://localhost:18966", 
        width       = 500, 
        height      = 300, 
        resizable   = true, 
        debug       = true,
        handler     = proc (w: Webview, arg: cstring) =
            echo "Got: " & $(arg)
    )
    wv.run()
    wv.exit()

proc initServer() =
    echo "server thread: " & $(getThreadId())
    proc main {.async.} =
        var server = newAsyncHttpServer()
        proc cb(req: Request) {.async.} =
            echo (req.reqMethod, req.url, req.headers)
            let headers = {"Content-type": "text/html; charset=utf-8"}
            if req.url.path == "/":
                await req.respond(Http200, "Hello World <a href='/boom'>Go to boom</a>", headers.newHttpHeaders())
            else:
                await req.respond(Http200, "In boom <a href='/'>Go back</a>", headers.newHttpHeaders())

        server.listen(Port(18966))
        let port = server.getPort
        echo "server running at: localhost:" & $port.uint16
        while true:
            if server.shouldAcceptRequest():
                await server.acceptRequest(cb)
            else:
                await sleepAsync(500)

    waitFor main()

spawn initServer()

initWebview()
```

PR #423, opened 2022-02-04
Convert all appropriate `proc`s to `func`s

drkameleon:
# Description

As per the latest Nim releases, a `proc` without side-effects (that is: `proc ... {.noSideEffect.}`) would have to be marked as a `func` ("pure" function).

So, let's go for it. 😄 

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge 🚀

Issue #385, opened 2021-12-01
Windows .cmd installer for self-building


Kyllingene:
**Is your feature request related to a problem? Please describe.**
Attempting to build arturo on Windows is painful, because there is no `install` script for Windows and WSL has some problems with it.

**Describe the solution you'd like**
A Windows `install.cmd` or something equivalent would streamline the building process for Windows tremendously.

**Describe alternatives you've considered**
Somehow altering the existing `install` script to work with WSL would fix the issue as well, but I'm not sure if that's possible.

I attempted to build with Nim, but ran into issues (particularly with a missing `gmp.h`); making it easier to build with Nim directly could also help.

**Additional context**
I would use the `curl` method of installation, but something is wrong with it on my machine at the moment; either way, being able to build on Windows would be helpful.

drkameleon:
Hi!

And thanks a lot for the suggestion!

I admit the support for Windows is not as complete as it should be - the main reason being that I'm totally not a Windows developer or even familiar with the OS (as a matter of fact, I don't think I even have a machine running Windows right now, but that could be solved somehow).

In any case, I'll definitely look into it, as it's important for everyone to be able to build/install Arturo on any system, in the easiest possible way. (Any ideas/pointers would be more than welcome!)

In the meantime, I could point you to the Windows CI build workflow (the one running automatically on a Windows server, every time there is a new push to the repo) - it's definitely not been easy to write all this and make it work, but it sure *does* work. So... you may get an idea of how this could work on Windows:

https://github.com/arturo-lang/arturo/blob/master/.github/workflows/windows.yml

P.S. Regarding the missing `gmp.h`, this is due to the fact that the "full" version requires the GMP (bignum) library to be installed. If I were you, I would first try building the MINI version (same as the normal one, minus a couple of not-so-commonly-used library, e.g. GMP): `./install mini`

Kyllingene:
Ah, ok; thanks! I understand that building for Windows as a Linux user is much harder than the other way around (due to how proprietary Windows is); I very much appreciate this language as a great alternative to other interpreted languages, and even more so the amount of effort you put into it.

drkameleon:
> I understand that building for Windows as a Linux user is much harder than the other way around (due to how proprietary Windows is)

[...]

PR #425, opened 2022-02-08
Clean up installation nimscript

drkameleon:
# Description

Thanks to @skydive241, we now have a NimScript-based installer (that can also be used for building, packaging or even running Arturo's unit-tests from one place).

Let's try to clean it up a bit and make it work in an identical way as the older install script (but without the mess there was 😄 )

Fixes: #293 #310 #385 #424

## Type of change

- [X] Code cleanup


skydive241:
One comment from my side as a windows user....

Windows doesn't care about shebang lines - which program/compiler/interpreter or whatever is responsible for a specific file ONLY depends on the file's extension (that's the reason why I called the file "install.nims", where the part "nims" tells windows to use nim.exe if I double-click on this file).

Windows users normally "fail to see" files without extensions,,,

Greetings
Patrick

drkameleon:
> One comment from my side as a windows user....
> 
> Windows doesn't care about shebang lines - which program/compiler/interpreter or whatever is responsible for a specific file ONLY depends on the file's extension (that's the reason why I called the file "install.nims", where the part "nims" tells windows to use nim.exe if I double-click on this file).
> 
> Windows users normally "fail to see" files without extensions,,,
> 
> Greetings Patrick

Just on time! 😄 

I wanted to ask you since I'm - quite obviously - completely lost when it comes to Windows.

First of all, let me tell you that you've done an amazing job. The script seemingly has undergone a load different changes since then (mainly, clean up my own code, make use of the fact that we are using Nim - and not Bash - as the script's language, and - my only "issue" - get rid of the Nim's "task" system: being a Unix user, being forced to write it like `nim install installer.nims mini` seems quite messy), the main basis of this change was the code you sent me (and you definitely made my life so much easier because with all honesty I'm not so familiar with NimScript or its specific commands).

That being said, the name of the file (and the shebangs, etc) has been bothering me.

As a Unix/Mac user, I'd definitely go for the shorter name (simply because `./build install` looks tighter and easier to memorize), but I do get your point, so I'm definitely willing to add the extension.

The question is: given that the script right now is to be run like: `./build.nims install`, how would double-clicking on it work? (I mean, given that it will definitely need extra arguments).

Is there anything like shortcut Windows scripts where we could have something click-and-run, translated to whatever we want? E.g. `nim build.nims install` or whatever for that matter.

Supplying a similar, tiny extra "executable" script for Windows only would solve all issues instantly.

Let me know and I'll proceed! ;-)

drkameleon:
As per the initial issue (https://github.com/arturo-lang/arturo/issues/385#issue-1068972232), would it make sense if we kept things as is and add e.g. a `install.cmd` script with something like that? (purely guesswork, I'll definitely have to create some Windows VM to try this out...)

```
nim e build install nogmp
```

skydive241:
Yes, you are right - double-clicking doesn't make sense if you need arguments put to the script. And I think, even windows users use a shell for those "building tasks". Using nim as interpreter, arguments could be passed by config.nims for example, and of course, links with arguments to doubleclick are also possible. 

Using file extensions simply is what windows users are used to: windows shows the appropriate icon next to the file name e.g. in file explorer, and windows knows which program to use IF you double-click the file (the reason why your arturo scripts should use the extension .art). Nim e.g. knows that ".nims" files are scripts to execute, and ".nim" files are programs which have to be compiled. Starting a file without extension needs to use a complete command line similar to the shebang: "nim.exe e build install".

To make a long story short: Using a file extension isn't necessary, but windows users are used to. 😂😂😂

drkameleon:
> Yes, you are right - double-clicking doesn't make sense if you need arguments put to the script. And I think, even windows users use a shell for those "building tasks". Using nim as interpreter, arguments could be passed by config.nims for example, and of course, links with arguments to doubleclick are also possible.
> 
> Using file extensions simply is what windows users are used to: windows shows the appropriate icon next to the file name e.g. in file explorer, and windows knows which program to use IF you double-click the file (the reason why your arturo scripts should use the extension .art). Nim e.g. knows that ".nims" files are scripts to execute, and ".nim" files are programs which have to be compiled. Starting a file without extension needs to use a complete command line similar to the shebang: "nim.exe e build install".
> 
> To make a long story short: Using a file extension isn't necessary, but windows users are used to. 😂😂😂

OK. You convinced me! Haha. 

The file will have its extension back. Now, I have no idea how things are going to turn out - I'll make a couple of changes here and there and attempt to integrate into the CI builds (let's how messy it becomes lol).

P.S. Regarding the `config.nims` you mentioned, to be honest I'm quite scared of them... Nim use one config file on top of another from different locations, and - unless you know *exactly* which ones are being used and/or explicitly tell the compiler to ignore some of them - I usually don't know what the final compiler arguments are. But, let's see.... I'm feeling adventurous today :p

skydive241:
> P.S. Regarding the config.nims you mentioned, to be honest I'm quite scared of them... Nim tends to use one config file on top of another from different locations, and - unless you know exactly which ones are being used and/or explicitly tell the compiler to ignore some of them - I usually don't know what the final compiler arguments are. But, let's see.... I'm feeling adventurous today :p

Yes, you are right - I needed DAYS to understand nim's "starting scripts" system, and usually, windows users work with shells like linux users do, I think. Please DO NOT create one more starting script like install.cmd or similar - those who care about building arturo WILL understand your nimscript file, regardless with or without extension. 😊😊😊

drkameleon:
> > P.S. Regarding the config.nims you mentioned, to be honest I'm quite scared of them... Nim tends to use one config file on top of another from different locations, and - unless you know exactly which ones are being used and/or explicitly tell the compiler to ignore some of them - I usually don't know what the final compiler arguments are. But, let's see.... I'm feeling adventurous today :p
> 
> Yes, you are right - I needed DAYS to understand nim's "starting scripts" system, and usually, windows users work with shells like linux users do, I think. Please DO NOT create one more starting script like install.cmd or similar - those who care about building arturo WILL understand your nimscript file, regardless with or without extension. 😊😊😊

Awesome! Working on it! 

I'll start pushing the updated CI builds in 5 minutes - and fingers crossed! 😄

drkameleon:
**UPDATE:** We're getting there. After a lot of efforts, I managed to find out what was going on with failed Linux builds and it was none other the "binary compression" part (apparently, [`upx`](https://upx.github.io/) ends up destroying our *nix binaries). Meaning: we'll either have to completely forget about it (for now, I have just commented out the related part for non-WEB builds) or find a solution -- not that it will matter so much, given that they new "package" function can already produce slim, tight and portable binaries (at least, much more than before).

That being said and seeing that the CI builds finally work fine 🤞, I'll leave things as-is and revisit the PR tomorrow, with a fresh set of eyes, to see if I have missed anything. 

If all goes well, the PR will be "officially" merged by tomorrow 😉

drkameleon:
Ready to merge 🚀

PR #443, opened 2022-02-21
Slimmer, more-compressed web version

drkameleon:
# Description

Currently, the original version of Arturo.js is around 3952524 bytes, while the minified one (with *uglifyjs*) comes down to 1171181 bytes (which is still quite a lot, especially if this was ever to be used in an actual website).

That being said and looking at the produced code (even when minimized) there seem to be many ways this final .js code can be even tighter. So, let's go for it.

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge 🚀 

-------

The PR doesn't make a *huge* difference, but still we managed to get a `1171181` -> `941559` decrease in size (roughly ~20%), which in turn is translated into **187154 bytes** after tar-gzipping. Not bad at all, for practically an entire programming language with a VM.

PR #444, opened 2022-02-25
Better System\execute function

drkameleon:
# Description

Currently, [System\execute](https://arturo-lang.io/documentation/library/system/execute/) takes a shell command (as a `:string`) and returns its output, synchronously (again as a `:string`).

The possibilities here are endless; thus, it would be good to add different options to it:

- [x] Execute a command string (as before) but also - optionally - return its exit code
- [x] Execute a command without returning its output, but have it all show as usual
- [x] Execute a command with an optional array of "arguments"
- [x] Execute all of the above as a process (meaning: *asynchronously*)
- [x] In the above case, also - optionally - return details about the spawned process (e.g. its *id*)
- [x] Add library function (e.g. `terminate`) to support "kill"-ing active processes - owned or external - by PID 

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge 🚀

PR #447, opened 2022-02-28
Fix Reflection\info function

drkameleon:
# Description

Currently, [`info`](https://arturo-lang.io/documentation/library/reflection/info/) works fine - even for user-defined variables/functions/etc.

The problem is when the passed (user-)function in question doesn't have any arguments. Then, the function gets stuck.
The same goes for `info.get`

## Example

```red
f: function [][ print "Hello world!" ]

inspect info.get 'f
```

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge 🚀

PR #448, opened 2022-03-01
Fix functions with already-defined function param(s)

drkameleon:
# Description

We have to make this work properly:

```
$> f: function [print][inspect print]

$> f 10
10 :integer

$> print 10
10
```

**Meaning:** a function should be able to re-define any symbol (even if it's a built-in function), fix the *Arities* table and restore it after the function call.

## Type of change


drkameleon:
Ready to merge 🚀

Issue #458, opened 2022-03-08
Add ability to search across functions by descriptive string or types


asampal:
An apropos command would be a nice addition to the REPL. For getting the lay of the land, Arturo currently has `help` and `info`. The first is all-inclusive, showing all available functions, while the latter focuses on one, specifically. Sometimes you don't know exactly what you're looking for, but you know how the desired function might have been described.

As an enhancement, an apropos-type of lookup would take a string (or maybe regex) and search across all functions, returning those where there is a match in the description (perhaps example comments too?).

Potentially, a variant of apropos could take a set of types and return results based on parameters matching.

Here's Julia's "apropos" as an example:

![image](https://user-images.githubusercontent.com/281639/157276915-cf8f4360-92a5-42cf-9419-ad23d6138d47.png)

drkameleon:
Another super-interesting idea!

> Potentially, a variant of apropos could take a set of types and return results based on parameters matching.

I had thought of that part specifically, but your proposal is even more complete.

Noted! 😉

drkameleon:
So, I'm doing some pre-release cleanup (https://github.com/arturo-lang/arturo/pull/461) and I decided to make the REPL a tiny bit better as well... 😄 

Part of this transition was to actually remove the `help` command and convert it into a REPL-only function; and thus [written 100% in Arturo](https://github.com/arturo-lang/arturo/blob/general-cleanup/src/scripts/console.art#L109-L117).

The exact same thing happened with your suggestion, which became [a new REPL-only function](https://github.com/arturo-lang/arturo/blob/general-cleanup/src/scripts/console.art#L119-L175) (`lookup`); with which we can now look up any built-in function either **(a)** given a string (where you we try to find it in the function name, function description, attribute name, or any of the attribute descriptions), or **(b)** given a type (where you try to find function with one of the parameters coinciding, or - with the option `.returns` - in its returned arguments).

Here's what the whole thing currently looks like:

### Our new `help` function

<img width="1023" alt="Screenshot 2022-03-10 at 12 41 52" src="https://user-images.githubusercontent.com/1265028/157655866-6d95a050-f4d4-4f1f-959a-499445cd72c1.png">

### The brand-new `lookup` function, in action

<img width="721" alt="Screenshot 2022-03-10 at 12 42 12" src="https://user-images.githubusercontent.com/1265028/157655933-7f8b3ae7-a357-4128-a5b3-9b76520f7d57.png">

#### Looking up by some string

[...]

PR #465, opened 2022-03-11
Add regex core type

### LINKED ISSUE #464: Add Regex core type?
See: https://github.com/arturo-lang/arturo/issues/458#issuecomment-1064462460

drkameleon:
# Description

As stated [here](https://github.com/arturo-lang/arturo/issues/458#issuecomment-1064904272) as well, there is already some extensive use of regular expressions throughout Arturo's standard library - exclusively via the use of a `.regex` option, where the given string/pattern is to be interpreted as a regular expression.

So... given Arturo's inherent type-flexibility, we could obviously add an extra data type, specifically for regular expressions.

### Now, how will we define them?

**One idea would be to do it with a converter** (pretty much like [`function`](https://arturo-lang.io/documentation/library/converters/function/), [`array`](https://arturo-lang.io/documentation/library/converters/array/) or [`dictionary`](https://arturo-lang.io/documentation/library/converters/dictionary/)), so that something like `regex {myregex}` works - and then probably add an alias to this new `regex` function. But this doesn't look too good simply because, the difference between `regex` and the existing situation (a `.regex` option) is pretty much trivial, plus we would have to pick a really short alias for that, and for that matter pretty much all the "good" ones have already been assigned to far more common operations.

**Another idea would be to do it via the parser**, that is: using some special syntax (pretty much like `"..."` is a *:string* and a character within backticks is a *:char*).

The obvious(?) approach then would be to consider `/myregex/` which is a rather common syntax. But then, this would lead to many different issues with the parser (since `/` is already a symbol, aliased to `div` - and, on top of it, without clear distinctive patterns).

### Conclusion

So... the idea we'll go for here is a hybrid of what has been mentioned:

Given that we already have syntax for "literal" strings, without character-escaping (`{...}` and `{:...:}`), we could simply add a different "subtype" of unescaped strings (which is actually what a regex is anyway), using the curly-brace outer wrapper and the common regex syntax mentioned above: 

like `{/this is my regex/}` 😄 

Obviously, this will be accompanied by the usual `to` utilities (e.g. `to :regex "some string"`) and will take the place of string arguments in functions that potentially expect a regular expression (= `.regex` option)

Fixes #464 

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

Issue #454, opened 2022-03-05
Symbols ending in `?` can't be jumped to in the documentation


asampal:
Due to the fact that `?` is a semantically important character in URLs, the wiki documentation has a problem jumping to symbols that end in `?`, e.g. `type?` and others.

It might be possible to escape this somehow to be able to achieve the hyperlink.

drkameleon:
The wiki documentation is quite an interesting issue.

Initially, the only documentation for Arturo was to be found at the wiki section - and auto-generated for that matter. Then I created the "official" website, with all the documentation in there, and in a far more efficient way IMHO, since there are many things that a GitHub wiki does not allow (e.g. proper navigation & syntax highlighting for languages not already supported by Linguist).

Right now, we are at an intermediate stage where - whenever I update the documentation - both sets (the one at the wiki + the one at the website) get auto-generated.

Truth be told, I was thinking of completely removing the wiki documentation and leave it all at the website. Or perhaps, keep doing what I do: have a "simplified", GitHub-friendly version here and one at the website. (+ I must mention that I need to clean up a bit the generator scripts - I mean for the documentation - but I've been putting off until the next release is about to come out...)

I'm a bit confused. Any idea is more than welcome! 😄

drkameleon:
I totally see your point.

Let me shed some light into the inner workings of the documentation (at least, as of now)...

We mainly have 2 sections:

- **The library** - with all the different modules + functions in each module, etc. The data for this comes directly from the source code. Have a look at the `get` function you mentioned for example: https://github.com/arturo-lang/arturo/blob/master/src/library/Collections.nim#L368-L404: All of this data is "embedded" into Arturo and of course remains accessible (you can pretty-print it in the REPL with *info*, e.g. `info 'print` or get it as a dictionary, e.g. `info.get 'print`_. So, in a few words, anybody can already contribute or fix anything related to library functions (even add examples as you can see). And when a new release is created, the main documentation site will be updated too.

- **The other pages** - these are basically the rest of the pages, mostly with text content. These are based on a hybrid between Arturo+Markdown (I've created a static-site generator in Arturo; only problem being, for now it works fine only with the documentation site lol) and you find them in the `/docs` folder. For example: https://github.com/arturo-lang/arturo/tree/master/docs/website/pages/documentation

As there are other people that have asked me about contributing, and - of course - I'm more than open to it... I'd say I need a couple of days to look into the doc generator again and we will all be good to go.

[...]

PR #469, opened 2022-03-12
Documentation + Wiki cleanup

drkameleon:
# Description

Right now, the situation with the documentation & the wiki (@ github) is quite messy. Not in the sense of being incomplete (which I hope is not right, though there will always be margin for improvements); mostly in the sense, that it's quite all over the place:

- The main documentation is at the ["official" website](https://arturo-lang.io/documentation/); this works quite fine, with a couple of asterisks: 
    - [x] we have to make sure its generation (see: Webize) works flawlessly
    - [x] re-organize the templates/sources for that
    - [x] thoroughly document it (what is where, and where everything can be found - so that everyone can contribute at will)
    - [x] make the "Related" section work a tiny bit more... efficiently (right now, there are cases where it works fine, but there are other case where the suggested "related" functions have nothing to do with the main function whatsoever...)
- The wiki here at Github has to change radically. Basically, the wiki was conceived as an easy way of having the library documentation (auto-generated) in one place, *before* the site existed. Now that the site is there and working fine, there is no need for that, especially in its current form. And, obviously, there is no need to duplicate things, because it confuses more than it helps (see: https://github.com/arturo-lang/arturo/issues/454#issuecomment-1060751044). So:
    - [x] remove all current wiki content
    - [x] remove connected workflow
    - [x] clean up all generators related to the wiki
    - [x] move things that are a bit too generic and not so super-important, from the main README (e.g. "What editors to use?") to the Wiki
    - [x] re-structure it, so that it somewhat obvious where new stuff should go
    
So, let's go for it. Arturo v/0.9.80 is too close. 🎸 

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #474, opened 2022-04-13
Add support for Rationals

drkameleon:
# Description

Pretty much as it happens with complex numbers (`:complex`), it would be nice to also have built-in support for rational value types (`:rational`). They are not going to come with any special syntax (as complex numbers), but they will be constructible via the `to` function.

## Type of change


drkameleon:
Ready to merge 🚀

PR #495, opened 2022-07-05
Combinatorics-related standard library functions

drkameleon:
# Description

Right now, we have `permutate` in the Collections module. 

**What has to be done:**

- `permutate` currently works on the entire sequence, and without repetitions; by default. This could be far more flexible
- in the exact same fashion, we could add another function, only this time for combinations (e.g. `combine` - not that this word is free, since the older `combine` has been renamed to `couple`, to avoid confusion)

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge 🚀

PR #512, opened 2022-07-13
Create new Strings\alphabet built-in function

drkameleon:
# Description

This new function will take *one* parameter (locale: the language ISO-2 code) and will return the whole alphabet, dictionary-index style, as a `:char` block. It will also have a few options:

- `.all`: include characters that *may* be encountered in the language, but do not form part of the language's dictionary index (e.g. the spanish character `ñ` is both a character in the Spanish language and would be used in a dictionary as an index; the character `á` on the other hand, although it may be part of a Spanish word, it does not constitute a valid index -> this one would be returned only when using `.all`)
- `.upper`: return the charset in UPPERcase

## Type of change


drkameleon:
Here's an example of the function in action (complete, with all possible option combinations), for Spanish (**es**):

```red
alphabet 'es
; => [a b c d e f g h i j k l m n ñ o p q r s t u v w x y z]

alphabet.upper 'es
; => [A B C D E F G H I J K L M N Ñ O P Q R S T U V W X Y Z]

alphabet.lower.upper 'es
; => [a b c d e f g h i j k l m n ñ o p q r s t u v w x y z A B C D E F G H I J K L M N Ñ O P Q R S T U V W X Y Z]

alphabet.all 'es
; => [a b c d e f g h i j k l m n ñ o p q r s t u v w x y z á é í ó ú ü]

alphabet.all.upper 'es
; => [A B C D E F G H I J K L M N Ñ O P Q R S T U V W X Y Z Á É Í Ó Ú Ü]

alphabet.all.lower.upper 'es
; => [a b c d e f g h i j k l m n ñ o p q r s t u v w x y z á é í ó ú ü A B C D E F G H I J K L M N Ñ O P Q R S T U V W X Y Z Á É Í Ó Ú Ü]

PR #563, opened 2022-07-15
Make dictionary functions callable directly 

drkameleon:
# Description

Arturo supports dictionaries with keys and values of any type.

A normal use-case would be:

```red
person: #[
    name: "John"
    surname: "Doe"
]
print person\surname
```

In case we had a field with a `:function` value, this should be callable as well:

```red
pack: #[
     fun: function [x][
           x + 1
     ]
]

; we should be able to do the following, even if we have to enclose it in a do-block:
print pack\fun 2 ; should print 3
```

Currently, this is not working; but it should. (What we can already do is call the aforementioned nested function via `call`, e.g. `call pack\fun @[2]`, but it should be even more straightforward...)

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge 🚀

PR #564, opened 2022-07-21
Create new `:quantity` type

drkameleon:
# Description

OK, so the idea is simple... We'll be *attempting* to add a new compound built-in type that will be used for expressing quantities, that is: a numeric value + a unit.

The idea stems from Rebol's original [`money!`](http://www.rebol.com/r3/docs/datatypes/money.html) datatype, which is meant to hold currency values, only we're going to make it much more uniform & less rigid (as far as syntax goes), less US-specific and covering as many different number/unit combinations as possible.

### A few ideas

- Currencies (_USD, EUR, CAD_, etc)
- Length (_m, mm, cm, in_, etc)
- Area (_m2, ha_, etc)
- Volume (_m3, li_, etc)
- Weight
- Time
- Temperature (_C, F, K_, etc)

And perhaps even more...

The initially proposed syntax is rather simple:

```
NUMBER:UNIT
```

That is: a numeric (`:integer` or `:floating`) part, immediately followed by a colon (`:`) and a unit abbreviation. E.g.:

```
money: 100:USD
length: 30:in
```

### The final goal

...is to be able to handle quantities in a uniform way, pretty much as every other number (all arithmetic operations included) but taking into account the respective units.

For example:

- `100:USD + 200:USD` would yield `300:USD`
- `100:USD + 200:EUR` would yield the sum in _USD_, after having converted the `200:EUR` to _USD_ (pretty much what Google does automatically)
- `100:USD + 30:F` (where *30:F* is 30 degrees Fahrenheit) would be an impossible operation that would throw an error.

Of course, having the language support quantities with units, this would mean any conversion could be done automatically. 😄 

## Type of change

- [x] Code cleanup


celtic-coder:
Hi Yanis (@drkameleon),

Have you considered the [Frink](https://frinklang.org/) language for looking at units of measure? There is an associated [units.txt](https://frinklang.org/frinkdata/units.txt) file that is both mind-blowing and humorous! Look out for the "Alan's editorializing:" comments for some biting sarcasm on what is sometimes seen as a boring topic!

Kind Regards,
Liam

drkameleon:
@celtic-coder Hi! 😄 

That sounds like a great idea. Thanks!

I've had a brief look into the project (and, yep, I *love* coming across programmers' gems hidden in the code lol) and it's definitely interesting. I guess that will be my main reference (especially for abbreviations), since a lot of the background work seems to have already been done.

This PR started a few weeks ago... in my head, as something that *could* be done. I'm still not sure how this whole thing will turn out, but (over)stretching Rebol's currency idea seems like something that (a) wouldn't look too bad in Arturo, syntax-wise, (b) can work smoothly with many already-existing builtin functions in our library, (c) would vastly simplify something that is needed quite often, in a natural way, and quite seamlessle integrated with the rest of the language. So... I decided to, at least, give it a go.

Let's see, let's see...

Thanks again for the suggestion - looks very useful! (Needless to say, any other suggestion is totally welcome! :wink)

drkameleon:
A sneak peek of our early progress (just before calling it a day, for today...):

Screenshot 2022-07-21 at 18 39 43

EMPAT94:
Hey there. 

Just found out about Arturo today! Appreciate the awesome documentation efforts; and the langugage is just lovely! Looking forward to experimenting with it in coming weeks.

Regarding `:quantity` type: The idea is absolutely fantastic for physical measures and invariant conversions.

I am a bit hesitant about inclusion of currency in the mix, however. My 2 cents (no pun): 

First, the changes to the exchange rates are arbitrary and frequent. Which means by definition, the language itself will depend on external (unverified?) sources for source of truth. Not a point in favor of reliability.

Second, it will necessarily require a network request. What happens when connection is poor? source is blocked by govt? requires proxy to access? What if there is no internet at all? These concerns are just top off my head.

Those are some of my concerns. I'm sure you've thought of them and more already, so I'm curious to hear your thoughts on the matter. Cheers!

drkameleon:
> Hey there.
> 
> Just found out about Arturo today! Appreciate the awesome documentation efforts; and the langugage is just lovely! Looking forward to experimenting with it in coming weeks.
> 
> Regarding `:quantity` type: The idea is absolutely fantastic for physical measures and invariant conversions.
> 
> I am a bit hesitant about inclusion of currency in the mix, however. My 2 cents (no pun):
> 
> First, the changes to the exchange rates are arbitrary and frequent. Which means by definition, the language itself will depend on external (unverified?) sources for source of truth. Not a point in favor of reliability.
> 
> Second, it will necessarily require a network request. What happens when connection is poor? source is blocked by govt? requires proxy to access? What if there is no internet at all? These concerns are just top off my head.
> 
> Those are some of my concerns. I'm sure you've thought of them and more already, so I'm curious to hear your thoughts on the matter. Cheers!

Patesh, hi and welcome! :)

And, of course, thanks A LOT for your kind words - I'm doing my best! 

Regarding the `:quantity` type, as I said, my main idea was to extend the Rebol idea (which was solely limited to money/currency) and add different other units that suit this format, so that we can handle quantities as a native type, that is: as part of the language.

Conversions came as an afterthought, since we'll obviously need to perform - at least - the basic arithmetic operations on these new values, so what would happen if they are of the same "kind" (e.g two different length units), but we want to add them up?

So, basically it was like a natural follow-up to the original idea.

Regarding your concerns (which are totally valid and, yes, I have thought about it too...):

- we should, I guess, explicitly state in the documentation the volatile aspect of this feature (that obviously rates change all the time) and whatever the shortcomings of this could be

- in case at some given moment an internet connection is not available, it could throw an exception (which we could catch with a `try` statement); I have also thought of storing perhaps the existing rates per day, so that we don't have to retrieve them for every single request/operation

- regarding the reliability of the source for the rates, I have thought of "us" becoming the source. That is: arturo-lang will be providing it in a semi-official way.

Of course, I still get your point, but a) it's something related to only one part of this whole new feature, b) the feature in itself is ultra-useful (I cannot cound the number of times I've needed something similar - personally I would say... on a daily basis), c) provided that the limitations/caveats are clearly stated, the benefits of treating all `:quantity` values in the exact same way, while still "hiding" the underlying complexing, I think they clearly outweigh the problematic parts...

Ofc, any input on the above is more than welcome! 😉

EMPAT94:
> the feature in itself is ultra-useful

I can already imagine using it regularly 👍🏼 

Explicit warnings in documentations and throwing error when no internets seems to be the way to go.

> arturo-lang will be providing it in a semi-official way.

Would there be any way to override the rates? 

---

On a side note, is there a way of importing one local script into another? 

To give a sample scenario, say I have a script `notify.art` that just pings me on a slack wehbook, sends a email or some such.

Would it be possible to import it into say, `monitor-website-one.art` and `check-available-memory.art` ? 

I tried searching but couldn't find any references. I'm sorry if this isn't the right place to ask. Let me know, I'll edit it out right away.

Thanks!

drkameleon:
> > the feature in itself is ultra-useful
> 
> I can already imagine using it regularly 👍🏼
> 
> Explicit warnings in documentation and throwing error when no internet seems to be the way to go.
> 

🚀 

> > arturo-lang will be providing it in a semi-official way.
> 
> Would there be any way to override the rates?

That's something that could definitely be considered. I'm all for flexibility and customisation, provided that we don't end up adding too much complexity. Let's see... Truth be told the hold concept is rather fresh, so I guess there are many things left to be figured out along the way...

> On a side note, is there a way of importing one local script into another?
> 
> To give a sample scenario, say I have a script `notify.art` that just pings me on a slack wehbook, sends a email or some such.
> 
> Would it be possible to import it into say, `monitor-website-one.art` and `check-available-memory.art` ?
> 
> I tried searching but couldn't find any references. I'm sorry if this isn't the right place to ask. Let me know, I'll edit it out right away.
> 
> Thanks!
> 
> Update: Found the `do.import {}` pragma from other projects. Imho, might as well add it in docs.

First off, `do.import "your-script.art"` *does* work fine for this exact thing. So, you're 100% correct.

Now, the reason why it's not been properly documented anywhere is that... I'm currently re-thinking and re-designing the whole package/module system, so that it works nice and in a uniform way. There are many different (hopefully minor) issues with that, so... I'm kindof taking my time until the whole thing is as it should be. (btw, I have already registered the `pkgr.art` domain name... so, you get an idea 😄 )

My goal is to make this as seamless as possible! Including scripts/including + auto-installing external packages/packaging your scripts in a standalone way, and much more.

So, I guess we'll all have to be just a bit patient... 😉

drkameleon:
Ready to merge 🚀

PR #572, opened 2022-07-26
Add support for big `:floating` numbers + GMP integration cleanup

### LINKED ISSUE #567: Floating-Integer (bignum) division not working properly
**Describe the bug**
Something like this: `2.1 / (5 ^ 31)` yields erroneous results.

A probably cause would be that `5 ^ 31` is already a GMP bignum and thus the division of a `:floating` value with a bignum-type `:integer` is not working as expected.

More here: https://discord.com/channels/765519132186640445/829324913097048065/1000934950767439973

drkameleon:
# Description

We could make use of GMP's `mpz_f` type and add support for "big" floating-point numbers, in the same fashion as we do with "big" integers (without having any distinct type *externally*, but actually using GMP's `mpz_t`s *internally*)

Fixes #567

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge 🚀

PR #576, opened 2022-07-30
Better unicode sorting

drkameleon:
# Description

Right now, we have unicode, locale-specific sorting (via [`sort`](https://arturo-lang.io/documentation/library/collections/sort/)) but it's just limited to Spanish (es) and English (en) - which is as if we dind't have it at all.

Given that we already have a charset helper (for now, only used in String\alphabet built-in function), we can combine the two and make a far better locale-aware sorting function. And without too much code repetitions. :)

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge 🚀

PR #577, opened 2022-08-02
Add support for digraphs/trigraphs (sorting)

drkameleon:
# Description

Right now, we have quite decent support for Unicode strings, sorting for various languages (using `sort`) and getting the charsets for each of these languages (via `alphabet`).

The problem is: there are - quite a few as a matter of fact - languages that have "digraphs". Digraphs/Trigraphs are actually 2- or 3-letter combinations that are regarded as ONE "letter".

A good example of this is Hungarian:

A | Á | B | C | Cs | D | Dz | Dzs | E | É | F | G | Gy | H | I | Í | J | K | L | Ly | M | N
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
Ny | O | Ó | Ö | Ő | P | Q | R | S | Sz | T | Ty | U | Ú | Ü | Ű | V | W | X | Y | Z | Zs

In this case, a word beginning with "cs" for example would be indexed AFTER a word beginning with "cu" (just because "cs" is a digraph that goes after "d" altogether)

That being said, the only way to add support for these languages as well would be to add support for di-/tri-graphs.

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #585, opened 2022-08-03
Convert custom-type user objects internally to distinct type

drkameleon:
# Description

Right now, objects can be created using a user-defined type, with [`define`](https://arturo-lang.io/documentation/library/converters/define/) (to define the new type) and [`to`](https://arturo-lang.io/documentation/library/converters/to/) (to actually convert a block to an object of the newly-defined type):

```red
define :person [name][] 
; we define a new type :person, 
; with one field -> "name"

me: to :person ["John"] 
; we create a new object of type :person, 
; with the name set to "John"
```

For now, these objects are internally `:dictionary` values, with a twist. It would be much (much) better if they were of a distinct `:object` type, so that we can generally write code in a more natural way - especially regarding object creation/"inheritance"/etc. (although Arturo will never be properly-speaking an OOP language, nor has it ever been the goal of this project)

Fixes: #653 

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #663, opened 2022-08-05
Create "real" benchmarking tool

drkameleon:
# Description

The more the project grows, the less easy it is to track down its performance; especially after major - or minor - changes.

My main model here is (paradoxically?) V: https://fast.vlang.io/

I *totally* love what they're doing there... not because any such number is as crucial as it may seem, but because you can track performance - even though it might be *roughly* - throughout the project's development circle.

So... the idea in some vague bullet points:

- Whenever a new commit/push happens to the master branch, after (and only if) the main Linux/full CI tests have passed, proceed to run them again - only with a different script that actually takes notes of the times for each one of them 
- Push the results to some other repo, only for the benchmarks
- Create a page on our official website that will showcase these results (preferrably, one per day - the last one)

----

## Potential issues

**Github CI services may end being not-the-most-appropriate solution to this type of problem** 
_In that case, I'd use a DO droplet_

**What if more examples/tests are added** 
_In that case, I guess we'd have to consider it a "milestone", so that results only after a specific milestone are taken into account_

**What time are we to take into account?** 
_I'd say the real/clock time, preferrably of the whole suite of tests being run, and prerrable twice so that we average the results_


drkameleon:
Ready to merge! 🚀 

----

P.S. The website part of the benchmark tool will be implemented in relation to the website/documentation. 😉

drkameleon:
⚠️ Completely superseded by https://github.com/arturo-lang/benchmarks

PR #691, opened 2022-08-25
Add Bytecode support for executable blocks + minor enhancements

drkameleon:
# Description

Arturo's standard library already has many functions that support *executable* blocks - that is: functions that take a block as a parameter, that is to be later evaluated (converted to bytecode) and then executed.

A quite obvious case of this are e.g. the functions in the Iterators module: `loop` takes - for example - a range and a param and then executes the given block as many times as needed.

Internally, that means the block is *first* evaluated to bytecode and then executed.

Obviously, if the loop is to be executed 1 million times, the aforementioned block will not be re-evaluated 1 million times. That's totally unnecessary. It's actually evaluated *once* and then executed accordingly.

However, let's take this as an example:

```red
loop 1..1000 'x [
    loop 1..1000 'y [
        print [x y]
    ]
]
```

This code has a couple of issues:
- The outer block will be evaluated once
- The inner block will be evaluated 1000 times (for every *x* in `1..1000`)

Let's say we could do this:

```red

action: to :bytecode [
    print [x y]
]
loop 1..1000 'x [
    loop 1..1000 'y action
]
```

This - if done on demand - could definitely speed things up.

And, in any case, even if this doesn't look great per-se (which it doesn't IMHO), why not just add some support for `:bytecode` values where the already-existing `:block` parameter will actually be evaluated? :-)

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #693, opened 2022-08-26
Better handiling of table query+access

drkameleon:
# Description

In Arturo's codebase, there are several cases where we want to check if a Nim table contains a specific key and - if so - do something with the corresponding value.

Right now, we're normally doing it like this:

```nim
if table.hasKey(someKey):
    let val = table[someKey]
    # and do something with val
```

This requires searching in the table twice (or even more times) and thus quite expensive compare to the simpler:

```nim
let val = table.geOrDefault(someKey, VNULL)
if val != VNULL:
    # do something with val
```

All possible instances of this structure have to be changed.

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #701, opened 2022-09-03
Add option for compressed bytecode & revise Arturo's compression capabilities

drkameleon:
# Description

- We should add the option (or not even as an option) of storing/retrieving bytecode in a compressed format
- Also, we could see whether MINI builds should be deprived of this feature
- Last but not least, see if replacing *miniz* with Zippy makes sense

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #708, opened 2022-09-05
Remove Zippy + add missing functionality to Miniz

### LINKED ISSUE #705: [VM/bytecode] replace zippy with already existing functions?
[VM/bytecode] replace zippy with already existing functions?

we could somehow use some of the existing miniz functions, to avoid the extra dependency

https://github.com/arturo-lang/arturo/blob/f194d4353079ce45693f443e9c9ec43f62c2fc74/src/vm/bytecode.nim#L16

```text
    import streams

    # TODO(VM/bytecode) replace zippy with already existing functions?
    #  we could somehow use some of the existing miniz functions, to avoid the extra dependency
    #  labels: vm, bytecode, enhancement, performance
    import extras/zippy

import os

# when not defined(NOUNZIP):
#     import extras/miniz

import helpers/bytes as bytesHelper

```

165f61c38ddee10401677fe8ce026b9c7c93c435

drkameleon:
# Description

No need to have a whole extra library (zippy), only to perform part of what Miniz - our already existing compression library - should be able to perform.

So, in a few words, we should add support for string/bytearray compression/decompression solely with Miniz. :)

Fixes #705

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #721, opened 2022-09-15
improve `cleanBlock`

### LINKED ISSUE #668: [VM/values/value] `cleanBlock` is too slow
[VM/values/value] `cleanBlock` is too slow

when built without NOERRORLINES - which is our normal setup - this specific piece of code could be slowing down the whole language by up to 20%

https://github.com/arturo-lang/arturo/blob/bc5bd6b2c8cb65042d1b80d140ce8d2c5431977a/src/vm/values/value.nim#L853

```text
        return x.params.a.len

# TODO(VM/values/value) `cleanBlock` is too slow
#  when built without NOERRORLINES - which is our normal setup - this specific piece of code could be slowing down the whole language by up to 20%
#  labels: vm, values, performance, enhancement, benchmark, critical
template cleanBlock*(va: ValueArray, inplace: bool = false): untyped =
        when inplace:
ew file mode 100755
ndex 000000000..6b871481e
++ b/tools/benchmark.art
```

683e44ff95214bc935542c0792abe238fb88a8bf

drkameleon:
# Description

Currently, `cleanBlock` from VM/values/value is a clear performance bottleneck: used in many, many different cases to filter Newlines from given a given block (used solely for debugging purposes & error reporting) it causes a substantial slowdown for anything block-related (or not so block-related).

In release/portable packages, or bytecode, this wouldn't be the case. But it *is* when running Arturo normally - which should be considered as the most normal scenario.

So... let's speed this up! :)

Fixes #668 

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #724, opened 2022-09-20
Replace internal `VNULL`s with *nil* references

drkameleon:
# Description

Internally, in many cases when e.g. we are checking for the existence (or not) of a specific symbol, attribute, etc, we are currently using `VNULL`s (which is a perfectly valid Value).

In this case, we have to perform equality tests (that test against this `VNULL`).

This type of tests are expensive (or at least more expensive than comparing to `nil`).

We should eliminate all such `VNULL`s throughout the codebase and replace them with simple `nil` references (so that Nim-style `.isNil` checks are made possible)

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #732, opened 2022-09-27
Better performance for Iterators

drkameleon:
# Description

The functions in the Iterators module are some of the most commonly used ones. Thus, given their rather-complicated implementation, they could be considered a performance bottleneck.

So... the module has to be optimized as much as possible.

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #745, opened 2022-10-06
Use `sink` constructors for complex values

drkameleon:
# Description

Many value constructors (e.g. `newString`, `newBlock`, etc), simply take the given param (e.g. a string) and encapsulate it in the form a Value object.

This - at times - leads to unnecessary value copies: why copy the string, when all we want is to create a value with it (e.g. `newString("hello")`, in that case `"hello"` should be directly passed to the object.
For that, we'll (try to) make some good use of [Nim's *sink* parameters](https://nim-lang.org/docs/destructors.html#lifetimeminustracking-hooks-nimeqsink-hook).

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #746, opened 2022-10-07
Eagerly evaluate Path blocks

drkameleon:
# Description

In the beginning was the Path.

And the Path looked something like this `arr\0\1`.

And this Path was evaluated to bytecode that looked something like this:

```
push 1
push 0
load arr
get
get
```

But what would happen if we had variable parameters? Easy, we can use blocks (e.g. `arr\[x]\[1+2]`).

The problem is such a path would lead to the exact same bytecode:

```
push [1 + 2]
push [x]
load arr
get 
get
```

And then **[`get`](https://arturo-lang.io/documentation/library/collections/get/)** would have to execute these two blocks to find out their result and *then* use it.

This is obviously a total overkill.

In the case of `Path` values (and `PathLabel` values too, ofc), we could simply evaluate passed blocks along with the path itself.

So that the path above (`arr\[x]\[1+2]`) would lead to something like that:

```
push 2
push 1
add
load x
load arr
get
get
```

This would avoid the extra block execution (with all their - current - overhead) and speed things up A LOT (especially the ones dealing with paths - in the case, e.g. for Rosetta, there are many, many uses; but, then, I think it's quite natural, and we'd rarely double-think about it before using something like `arr\[x+1]`)

So, let's go for it - it's totally worth the fuss!

## Type of change

- [x] Code cleanup


drkameleon:
Completed and rather easily, admittedly. 😄 

This PR gives us a rough 15% performance boost overall (with up to 3x faster runs - that is > 300%! - for code with heavy use of block assignment/access). As an illustrative example, Arturo's [implementation for **Van eck sequence** example from Rosetta Code](https://github.com/arturo-lang/arturo/blob/master/examples/rosetta/van%20eck%20sequence.art), on my Mac, dropped from ~9.5s to close to 2.9s (!).

So...

Ready to merge! 🚀

celtic-coder:
> Completed and rather easily, admittedly. 😄
> 
> This PR gives us a rough 15% performance boost overall (with up to 3x faster runs - that is > 300%! - for code with heavy use of block assignment/access). As an illustrative example, Arturo's [implementation for **Van eck sequence** example from Rosetta Code](https://github.com/arturo-lang/arturo/blob/master/examples/rosetta/van%20eck%20sequence.art), on my Mac, dropped from ~9.5s to close to 2.9s (!).

Good work! 😄

PR #747, opened 2022-10-07
Fix FreeBSD workflow and set Nim to 1.6.6

drkameleon:
# Description

See: https://forum.nim-lang.org/t/9510

## Type of change

### DIFF
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -46,11 +46,10 @@ jobs:
         id: test
         uses: vmactions/freebsd-vm@v0
         with:
+          release: 12.0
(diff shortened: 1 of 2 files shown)


drkameleon:
The main issue (moving back from Nim 1.6.8 to 1.6.6) seems to have been fixed by downgrading the host OS (FreeBSD) from 13.0 to 12.3.

However, we keep seeing the compiler getting "Killed" during compilation. 

I believe this has to do with template expander taking too long (given Arturo's code using templates extensively - which is another issue to be fixed later on?)

So... for now, let's consider this PR ready to be merged. 🚀

PR #750, opened 2022-10-07
WIP: Eliminate unnecessary copies from `for` loops

drkameleon:
# Description

There are many instances of a simple `for x in items` loop, triggering unnecessary copies for `x`.

This has to be fixed on a case-by-case basis.

## Type of change

- [x] Code cleanup


drkameleon:
Not stable. Possible regression. 💣

PR #755, opened 2022-10-12
Collections\append: Replace `for loop` by the optimized `cleanAppend`

### LINKED ISSUE #679: [Collections\append] Append for arrays vs strings seems to be much slower
[Collections\append] Append for arrays vs strings seems to be much slower

https://github.com/arturo-lang/benchmarks/blob/main/results/17-8-2022/micro.md

https://github.com/arturo-lang/arturo/blob/71347600dd16539298cb0d02e1303377821702d7/src/library/Collections.nim#L77

```text
        """:
            ##########################################################
            # TODO(Collections\append) Append for arrays vs strings seems to be much slower
            #  https://github.com/arturo-lang/benchmarks/blob/main/results/17-8-2022/micro.md
            #  labels: enhancement, library, performance
            if x.kind==Literal:
                if InPlace.kind==String:
                    if y.kind==String:
ndex 1665be972..4d171a548 100644
++ b/src/vm/values/value.nim
```

1efce2a6fadf66066c788219f42a2063e2cd451f

RickBarretto:
# Description

\*(***Needs Review***)\*

I replaced the for loop:
```nim
...
for item in cleanY:
    inPlaced.a.add(item)
...
```
by:
```nim
...
SetInPlace( newBlock( cleanAppend( InPlaced.a, y.a ) ) )
...
```

***I think it'll need a review***, *because I don't understand so much things occurring into Arturo.*
But, I hope it should fix.

If I'm not wrong:
- `SetInPlace` creates a new object, while `InPlace` just uses the already existing object
- So I created with `newBlock` a `cleanAppend` that is a optimized function to concatenate arrays
- And `obj.a` says that it's a array type, if I'm not wrong

Fixes #679

## Type of change

- [x] Performance fix (non-breaking change which fixes an issue)


RickBarretto:
@drkameleon , What have I done wrong?

RickBarretto:
I found the reason of fails, my automatic formatter is broking the code.

The problem given was a indentation error. When I write the code it's right, but when I check, its wrong.
I'll try to disable it.

RickBarretto:
Nope, It's broken... 😞 @drkameleon

drkameleon:
> Nope, It's broken... 😞 @drkameleon

First if all, no worries! :)

Now, I *think* using `InPlaced = ...`, instead of `SetInPlace` should fix the error you are getting. I'll look into it carefully tomorrow morning.

Now if you ask me, given that the the prmurpose of InPlaced is to actually change it internally, I think a second version of `cleanAppend` (taking a `var ValuaArray`) would be the way to go...

RickBarretto:
Just replace `SetInPlace(...)` with `InPlaced = ...` don't work. The CI is failing again.

src/library/Collections.nim
```diff
@@ -76,56 +76,55 @@ proc defineSymbols*() =
...
+                elif InPlaced.kind == Char:
+                    if y.kind == String:
                         SetInPlace(newString($(InPlaced.c) & y.s))
-                    elif y.kind==Char:
+                    elif y.kind == Char:
                         SetInPlace(newString($(InPlaced.c) & $(y.c)))
                 else:
-                    if y.kind==Block:
+                    if y.kind == Block:
                         ensureCleaned(y)
```
drkameleon, line 92:
```suggestion
```

We don't really need this (to pre-clean the `y` value that is), since `cleanAppend` is supposed to work on non-clean values after all.

src/library/Collections.nim
```diff
@@ -76,56 +76,55 @@ proc defineSymbols*() =
...
-                        for item in cleanY:
-                            InPlaced.a.add(item)
+                        InPlaced = newBlock(cleanAppend(InPlaced.a, y.a))
```
drkameleon, line 93:
```suggestion
                        # TODO(Collections\append) In-place appending should actually work in-place
                        #  labels: enhancement, library
                        InPlaced.a = cleanAppend(InPlaced.a, y.a)
```

`cleanAppend`, the way it's defined right now, takes two (not necessarily clean) ValueArray objects and returns a ValueArray. So, we can directly assign the result back to the `.a` field, which holds the ValueArray.

I'm adding the TODO since in-place methods are meant to work on the value itself.

Basically, here we are taking the symbol's values (`InPlaced.a`) and the second parameter's values (`y.a`), concatenate them together, get this news array and assign it back as the initial symbols values. So, technically, there is an extra array created, while we could have just appended the second parameter's values (`y.a`) to`InPlaced.a` directly.

For that we would need a 2nd version of `cleanAppend` working directly on the first parameter, e.g.:

```nim
proc cleanAppendInPlace*(dest: var ValueArray, src: ValueArray) = 
    # append src directly to dest
```

And just do `Inplaced.a.cleanAppendInPlace(y.a)` :)

src/library/Collections.nim
```diff
@@ -47,21 +47,21 @@ proc defineSymbols*() =
     #  labels: library, enhancement
 
     builtin "append",
-        alias       = doubleplus, 
-        rule        = InfixPrecedence,
+        alias = doubleplus,
```
drkameleon, line 50:
```suggestion
        alias       = doubleplus, 
```

Also, if you could restore the indentation in the built-in method headers, it would really appreciated! 😉  (This and all the other cases...)

drkameleon:
Generally speaking, the `InPlaced` manipulations are a bit tricky - implementation-wise - but they are totally worth the fuss.

Since many (if not most) of Arturo's built-in function work this way (with a literal/in-place) parameter, this can be very practical and efficient, for performance reasons.

What we're dealing with is...

**Normal appending**:

```rebol
a: [1 2 3]

a: a ++ [4]
a: a ++ [5]

; a => [1 2 3 4 5]
```

vs...

**In-place appending**:

```rebol
a: new [1 2 3]

'a ++ [4]
'a ++ [5]

; a => [1 2 3 4 5]
```

Actually the second case is far more performant (provided, obviously, that the internal implementation is). 😉 

A very brief overview of how the whole thing works:

- Each builtin function declared in the library has several specs (parameters, with specific types, etc)
- Each and every time a built-in function is called, based on these(^) specs the interpreter first verifies everything is right (correct number of arguments, with the required params, etc): https://github.com/arturo-lang/arturo/blob/master/src/vm/lib.nim#L174-L195
- If everything is right, the function values are auto-assigned, before the body is processed: `x` is the first value, `y` is the second one, `z` is the third one
- Values obviously are Arturo's values as defined as an internal object type here: https://github.com/arturo-lang/arturo/blob/master/src/vm/values/types.nim#L231-L325 (each one of a different "kind", and - based on this - with different fields of different types; if you have a good look there, it's quite easy to get an idea)
- Now, a built-in function, does one of the two things:
     - returns a value; by returning, we mean pushing a value onto the stack (all of the stack operations are defined here: https://github.com/arturo-lang/arturo/blob/master/src/vm/stack.nim, but I doubt you will ever need anything other than `push`
     - if the first parameter is a literal (and is meant to be processed in-place, like in the above example) it doesn't return anything, it just changes the symbol value internally. So let's say the literal value `'myvar` is passed as a first parameter. It is then assigned to `x` (as usual). But this `x` is nothing but a string (the name of the variable itself, not its content/value). We may then look it up in our global Symbol table (`Syms[x.s]`) or just use one of the templates/shortcuts, like `InPlaced`, all defined here: https://github.com/arturo-lang/arturo/blob/master/src/vm/globals.nim#L74-L90 - all these are nothing but Nim templates (mere code substitution to make the whole thing more readable nothing more).
     
Now, regarding this mess (because that's what it is) with `cleanedBlock` and the concept of "clean" blocks:

in normal builds, where we want to hold information for lines for error reporting, a parsed block may contain Newline values (that's how we keep track of where an error occured).

if you take this for example:

```rebol
x: [
   1
   2
   3
]
```
... and loop through it without "cleaning" it (removing the newlines), what you would be looping through wouldn't be `1,2,3` but something like `Newline,1,Newline,2,Newline,3,Newline` - which is obviously not what you would intend. That's where this concept of "clean" blocks comes into play. Obviously, blocks created by built-in functions are to be considered "clean", by definition.

Now, all that being said, I've been working a lot on these aspects, and it's not perfect yet by any means (so, different things related to that may change), but... that's a general overview.

In any case, a) good start - don't be discouraged at all! And b) don't hesitate to ask me for any clarification you may need. Once you get used to the general idea (Nim included), you'll realize it's not as intimidating as it initially looks! 😉

drkameleon:
P.S. Your PR made me re-visit a few things regarding clean blocks and looping through them. So, since Nim has a nice little thing called Iterators (basically, like templates that "yield" values), but in a zero-cost way, in order to avoid code repetitions, I've refined it like:

```nim
iterator cleanedBlockValues*(v: Value, L: int): lent Value =
        if v.dirty:
            for i in 0..L-1:
                if v.a[i].kind != Newline:
                    yield v.a[i]
        else:
            for i in 0..L-1:
                yield v.a[i]
    else:
        for i in 0..L-1:
            yield v.a[i]

template cleanedBlockValues*(v: Value): untyped =
    let l = v.a.len
    cleanedBlockValues(v, l)
```

and thus, this will be the new look of `cleanAppend` (now taking values, instead of arrays, since we want to be able to tell whether they are `.dirty` or not):

```nim
func cleanAppend*(s: Value, t: Value, singleValue: static bool = false): ValueArray {.inline,enforceNoRaises.} =
    let L1 = len(s.a)
    when not singleValue:
        let L2 = len(t.a)
        result = newSeq[Value](L1 + L2)
    else:
        result = newSeq[Value](L1 + 1)

    var cnt = 0
    for i in cleanedBlockValues(s, L1):
        result[cnt] = i
        inc cnt

    when not singleValue:
        for i in cleanedBlockValues(t, L2):
            result[cnt] = i
            inc cnt
    else:
        result[cnt] = t
        inc cnt

    setLen(result, cnt)
```

It still works on fixed block values (for in-place append, we'd still need an in-place variant) - but, the good news: even this slight change already gives us a tiny performance boost compared to the previous ones.

Long story short: I'll push the changes to the master branch - and if you want, you can follow from there... so as not to do the same things twice. 😉

drkameleon:
https://github.com/arturo-lang/arturo/commit/8eb898d201e935341c6e96d7f8003523613372d3

src/library/Collections.nim
RickBarretto: I have a doubt. What exactly is a clean value?

src/library/Collections.nim
RickBarretto: > Also, if you could restore the indentation in the built-in method headers, it would really appreciated! 😉 (This and all the other cases...)

I'll do it! 😉

src/library/Collections.nim
RickBarretto: You answered my question in [this comment](https://github.com/arturo-lang/arturo/pull/755#issuecomment-1277306758). Thank you!

Now it sees clear to me.

drkameleon:
@RickBarretto If you're done, I'll resolve the conflict it mentions & do some housekeeping to get the CI actions to run. Just let me know! :)

RickBarretto:
@drkameleon I restore the indentation.
The `.vscode/*` that I added into `.gitignore` was to not pull my settings that disables auto-formatting.

drkameleon (approved):
All good!

drkameleon:
And... @RickBarretto, you're the 2nd person ever to add/edit anything in Arturo's source code itself. (The other one is me 🤔 lol)

So... good job!

----

Ready to merge! 🚀

celtic-coder:
Good work on this, @RickBarretto ! 😄

RickBarretto:
@celtic-coder Thank you! 😉

PR #760, opened 2022-10-13
WIP: Convert unnecessary empty-value constructors to constant value copies

drkameleon:
# Description

Let's say we want to set a variable to an Logical-true Arturo value.

This:
```nim
let a = VTRUE
```

seems like 20% faster than this:

```nim
let a = Value(kind:Logical, b:True)
```

The same seems to be valid for all Value types (to be confirmed).

All these have to be fixed.

## Type of change

- [x] Code cleanup


drkameleon:
Marked as obsolete - there has to be a radical change concerning values & value copying; hence we'll work on it in a different PR. 🤔

PR #762, opened 2022-10-14
WIP: Create new values when pushing onto the stack

drkameleon:
# Description

...

## Type of change

- [x] Code cleanup
- [x] Bug fix (non-breaking change which fixes an issue)

PR #767, opened 2022-10-17
Introduce readonly values + Sanitize value copying & use of `new`

drkameleon:
# Description

Right now, there is a slight mess regarding value objects in Arturo, which may (or may not) also degrade general performance.

The most prominent example of the aforementioned... mess is the use of the function [`new` ](https://arturo-lang.io/documentation/library/core/new/)(clearly highlighted here: https://arturo-lang.io/documentation/language/#using-literals)

The usage outlined in `new`'s documentation is fine. The one outline in the language documentation (e.g. `new 1`) should be cleared treated as a bug. And, as such, should be fixed.

**Generally speaking:**

- we must clean up the way constant values are pushed onto the stack
- how passing values works internally
- be totally clear about when we are passing references or the values themselves
- keep use of `new` to a bare minimum

**Also related:** (failed attempts) #760 #762 

## Type of change

- [x] Code cleanup


drkameleon:
OK, so a little recap of what this apparently trivial (not at all) PR achieves:

**Constant vs non-constant values**

From now on, all values will be internally marked as constant/readonly or not. For example, `3` is a constant value. While in `x: 3`, `x` is *not* a constant value.

**Consistent implementation of value copies**

In Arturo, as a rule, values are passed by reference. So, whether you realize it or not, they are actually hidden pointers. (This makes things faster, in general, and practical).

In general, the idea is: when moving values around, it's always by reference. 

Now, in assignments, it's a tiny bit different:
- If the right-hand value (rvalue) is a constant one, then we are force to copy it
- Otherwise it, we still pass it by reference

**Normalized use of `new`**

The truth is I didn't need [Tomasz's bashing of the language](https://dev.to/taw/100-languages-speedrun-episode-91-arturo-29fh) to realize that Arturo's variables appeared too... "wild" indeed.

So, I just fixed this whole mess. 😄 

From now on:

- No need for things like `a: new 1` (even though we may modify `a` in-place later on)... since `1` is a constant value being assigned, the copy is implicit
- Symbol-to-symbol assignments remain as-is: `a: b`. If, later on, we in-place modify `a` or `b`, both will change. If you want to avoid that, we should actually assign a *copy* of `b`, which will happen as it already does (`a: new b`). NOTE: This rule is valid for any type of value - not just for arrays/blocks and dictionaries, as it happens in other languages (e.g. Ruby). Integers, for example, will behave in the exact same fashion, for reasons of consistency. 

----

P.S. In the meantime, all of the Rosetta Code examples that made too... much use of `new` have been adapted accordingly (now, let's see when I'll find the time to update them all properly back to RC... 🤔 )

drkameleon:
And...

Ready to merge! 🚀

PR #768, opened 2022-10-19
Replace all Syms access with `SetSym` & `GetSym`

drkameleon:
# Description

Right now, we are setting/getting different values in the global symbol table, either by directly accessing it (`Syms[something]=...`) or via `SetSyms`.

This makes the whole thing a bit confusing and more prone to errors.

By only and uniformly using the global accessors, we'll eliminate the risk of discrepancies or bugs, in the future.
 
## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #774, opened 2022-10-21
Remove duplicate `$` overload for Values

drkameleon:
# Description

Right now, we have 2 quasi-identical `$` overloads for Value objects.

One here: https://github.com/arturo-lang/arturo/blob/5afe934728a9677338e82cf503d2ed2b32659b79/src/vm/values/value.nim#L2010

And... one here: https://github.com/arturo-lang/arturo/blob/5afe934728a9677338e82cf503d2ed2b32659b79/src/vm/values/printable.nim#L37

The reason behind it is that printing objects calls exec, exec import values, and... bingo! Circular dependency.

However, since the issue of VM/values/value's `$` overload is apparently ultra-limited (and only relevant for that specific module), I guess we can clean this up like a lot.

Let's go for it! 😉 

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #781, opened 2022-10-24
Add `{.noInit.}` variables where appropriate

drkameleon:
# Description

Variables (especially, `seq`s and `table`s) that will be certainly initialized by the VM later on, shouldn't be previously initialized previously with any default value.

## Type of change

- [x] Code cleanup


drkameleon:
Regression! Aborted. 🧨

PR #782, opened 2022-10-25
Cleaner and more efficient scoping rules + various optimizations

drkameleon:
# Description

OK, so deep breath; this is going to be a big one.

The issue with scopes & scope handling in Arturo is one of its absolute flaws. (There, I said it 😢 ).
And it's not so much how it has been defined (*so far*) but mostly how it has been handled - internally.

In short: I believe it's responsible for a great deal of the performance bottleneck experienced when executing *any* type of Block and one bit of Arturo's internals that I'm not particularly proud of, implementation-wise.

So... 

Let's have a look at the current situation
====

Well, not much to say here... Every block creates (and, subsequently, destroys) a new **scope** (that will be our magic keyword for this PR). In a few words: do we declare a variable inside it? It always, no-matter-what  type of "always", stays inside. Trying to access it from the outside? That's an error. Does the variable already exist in the outer scope? It updates it.

And although all that sounds nice, and very C-ish, it has some severe drawbacks:

- The time needed for this storing and re-storing symbols, especially the way it has been done so far, takes ages (I'm not talking about a couple of iterations, I'm talking about the case of having *many*; and that's not such an exceptional case by any means...) 🐢 
- It creates issues when we actually want to "import" (or "export", depending on how we look at it) new variables to the outer scope; and thus leading to the necessity of littering the standard library functions with `.import` options and the like, just in case we end up needing to... 
- Most other interpreted languages right now - including the big ones like Python, JavaScript, etc - don't handle it like that all (OK, in that aspect, Ruby may be standing somewhere in the middle). But, even with asterisks, it seems as if Arturo has been doing it in its own, very unique and very... award-winning least-efficient way possible
- Arturo's heritage (see: Rebol) supports this so-called "declarative" scope throughout (and Arturo pretty much follows the paradigm, with a couple of exceptions I guess), thus... de facto going against the block-scoping idea

What this PR will attempt to do/change
====

Long story short: try to do it - at last - the right way

**New scoping rules:**

- Function blocks will be the only ones that will be retaining their scoping as-is.
- All other blocks will be executed *inline* (that is: without a scope).
    - Blocks taking injected parameters (all of the function in the Iterators module, e.g. [`loop`](https://arturo-lang.io/documentation/library/iterators/)) will also be executed scope-less, but preserving the parameters. In short: Do we loop with an `'x` parameter? This parameter will not leak to the outer scope, nor will it update an existing `x` variable - IF it existed
    - Dictionary or array blocks - which are already special anyway (see: `@[1 2 3]` or `#[user: drkameleon]`) will be treated in a special way

-----

All that having been said, after a couple of experiments, there are 2, very optimistic things I have noticed so far:
- 🕐 In certain, block-intensive cases, the performance gains - solely because of the aforementioned change - could be up to 25x faster, while the whole suite of Rosetta Code examples... even without any further optimizations, seems to be running at least 20% faster 
- 🦺 Despite the radical(?) change this PR is supposed to bring, of all the 480 tests running as part of the CI builds, only 1 broke (and that was the unit-test testing the scopes themselves; so, yes, that specific one was actually *meant* to break)  

So, so, so... let's rock'n'roll! 😄 🎸 

## Type of change

- [x] Code cleanup
- [x] Performance enhancement


celtic-coder:
That's a challenging piece of work, Yanis (@drkameleon)! But it will be worth it in the end, if all the gains are realised. Onwards and upwards! 😄

drkameleon:
> That's a challenging piece of work, Yanis (@drkameleon)! But it will be worth it in the end, if all the gains are realised. Onwards and upwards! 😄

Liam, thanks a lot for your kind words! :)

I do think it's worth it + I've been brainstorming different solution for the past months (literally)... so time to see them in action! (for now, all I can say is I'm more than impressed with the results - I'll post them as this PR comes closer to completion! 😉 )

drkameleon:
Ready to merge! 🚀

PR #784, opened 2022-10-28
Better Value type verification & identification

drkameleon:
# Description

In many cases, where we are trying to see if a value `v` is of a specific kind (or, better said, *one of* several given kinds), we use an expression like `v.kind in [Integer,Floating]`, or something like that.

This - instead of actually testing the kind - performs an array search, which is definitely costly.

All of the above references have to be replaced accordingly.

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #787, opened 2022-10-31
Misc. evaluator optimizations

drkameleon:
# Description

There are various things that can be improved in the Evaluator's implementation.

First of all, a very good clean-up.

Then, and that's a big deal - that should be thoroughly tested & benchmarked - many functions could be inlined during evaluation. E.g.: When we encounter something like `if X Y`, then both Y and X are pushed onto the stack, Y being a block, and then IF is called - as a normal stdlib function - and executes the block, or not, accordingly.

In that case, an - obvious - idea would be to inline it all, so that we don't move outside the interpreter loop. As far as I can tell, this should give us an important edge.

But again: after exhaustive testing, first.

So... let's see! :rocket:

## Type of change

- [x] Code cleanup
- [x] Performance enhancement


beef331:
I do not know if you use them much but all of [these](https://github.com/arturo-lang/arturo/blob/master/src/vm/values/custom/vbinary.nim#L30-L37) operations allocate twice for no reason. The following is a more memory friendly implementation
```nim
proc `and`*(a, b: VBinary): VBinary =
    result = newSeq[byte](max(a.len, b.len))
    for i in 0..result.high:
      if i in 0..min(a.high, b.high):
        result[i] = a[i] and b[
```

beef331:
Looking further [this block](https://github.com/arturo-lang/arturo/blob/master/src/vm/values/types.nim#L194-L211) should likely be turned into a `FuncType = ref object` to make the `Value` type smaller so you do not need to allocate around 13 64bit types for every `Value`. This will reduce memory consumption a bit and most likely increase performance a tinge.

drkameleon:
> I do not know if you use them much but all of [these](https://github.com/arturo-lang/arturo/blob/master/src/vm/values/custom/vbinary.nim#L30-L37) operations allocate twice for no reason. The following is a more memory friendly implementation
> 
> ```nim
> proc `and`*(a, b: VBinary): VBinary =
>     result = newSeq[byte](max(a.len, b.len))
>     for i in 0..min(a.high, b.high):
>       result[i] = a[i] and b[i]
> ```

First of all, hi! :)

I'm really glad that you've had a look into the source code - especially given your experience with Nim. (I'm pretty sure I have had more than one look into your repos - good job!)

The 3 functions in question (meaning: `and`, `or` & `xor` for `VBinary` values) are by no means very common or critical. BUT: I'm more than open to every fix or enhancement that is in the right direction. So, let's go for it!

If you want to make a PR, you are totally welcome to - and I'll merge it! 😉  (If not, I'll obviously fix it myself, based on your suggestion)

drkameleon:
> (Feel Free to tell me to sod off if you wish 😛 )

It's quite the opposite actually. It's exactly the type of input & suggestions that I need! 😄 

So, feel free to keep them coming. PRs are also welcome too, as I said, especially from someone with your Nim background!

Obviously, this doesn't mean that I'll accept *everything* or that all changes are that easy (I mean, some things regarding the core may involve a vast amount of changes throughout the code, so I don't know whether I would prioritize them), but they'll definitely find their place somewhere in the to-do queue...)

> Looking further [this block](https://github.com/arturo-lang/arturo/blob/master/src/vm/values/types.nim#L194-L211) should likely be turned into a `FuncType = ref object` to make the `Value` type smaller so you do not need to allocate around 13 64bit types for every `Value`. This will reduce memory consumption a bit and most likely increase performance a tinge. 

I know exactly what you mean. In the past few months (I'd say, 3-4), I've totally dedicated myself to iron out things that were "off" and especially performance-related issues. (The language was rather stable and I guess it was about time to start dealing with them as well - so far I'd say the VM has seen an average boost of 4-5x). The Value object is a key point as well. And the size of it, ofc. So, noted! 😉  

(I guess there are other parts of the object that make it more bloated than it should be... Let's see. Of course, this will involve killing myself with benchmarking - I think after this project I will be able to claim the title of professional, obsessive-compulsive benchmarker LOL)

drkameleon:
Since this PR has taken gigantic dimensions and I've pretty much lost count of the things included in here... I'm merging it so that we keep it a bit more organized.

There are of course things pending (including some further cleanup), but this will be dealt with by future PRs and enhancements.

Overall, the performance improvement expected solely from this PR is around 20-25% ⬆️ . That's more than good enough!

So... ready to merge! 🚀

PR #788, opened 2022-11-02
Change release builds configuration

drkameleon:
# Description

Right now, for the release builds - all debug symbols, etc are there making the final binary unnecessarily big.
Also, our MINI builds, apart from excluding all the libraries/modules/capabilities that are not meant to exist in the MINI builds, also compiles the whole project with `--opt:size`, which has little gains in size, but a more significant drop in speed.

Let's fix this! 😄 

## Type of change

- [x] Code cleanup
- [x] Performance enhancement


drkameleon:
Ready to merge! 🚀

PR #790, opened 2022-11-05
Rough pass at making Value smaller for possible optimisation

beef331:
# Description
This PR is mostly a test PR to see if there is any benefit in making the `Value` type smaller. It moves `Function`'s data to a `ref object` instead of being on `Value` which will reduce the size significantly for most allocations. All fields of `Function` have templates on `Value` to make it more seamless with the previous API.

## Type of change

- [x] Code cleanup


beef331:
I do not know how to use your `build.nims`. So I do not know how to benchmark locally. (Might I suggest making a `nimble` file so one can do `nimble build` and `nimble test`?)

PS:  
  > PRs are also welcome too, as I said, especially from someone with your Nim background!
  
Confuses me, do I have some sort of impressive background? I have 0 large project as you very much do.

drkameleon:
> I do not know how to use your `build.nims`. So I do not know how to benchmark locally. (Might I suggest making a `nimble` file so one can do `nimble build` and `nimble test`?)

Funny thing is Arturo *was* initially using Nimble, until I decided to take it to a new Nimble-independent direction.

The build script is simply used like: `./build.nims install` where it "installs" the resulting binary into the local `/bin` folder + @ `~/.arturo/bin`

To save with a different name, I do it like: `./build.nims install --as:arturo-X` and then use hyperfine for comparing 2 different versions.

For the whole CI suite to run, all you have to do is `arturo tools/tester.art`. it's good both for validating and measuring. And then, there is off the benchmarks repo (under arturo-lang), mostly with micro-benchmarks.
> Confuses me, do I have some sort of impressive background? I have 0 large project as you very much do.
Well, you most definitely have a decent amount of repos and most importantly around Nim - which I doubt anyone willing to contribute to Arturo has ever really had! :)

beef331:
Is there anyway to just run the few benchmarks and not all of them?

drkameleon:
> Is there anyway to just run the few benchmarks and not all of them?

The benefit of running the benchmarks for the entire set is that - especially for far-reaching changes - we can see the impact of any given change on a rather varied codebase.

Now, for more minor or more focused changes, what I normally do is:
- either create a tailor test (in Arturo) or pick one of the existing ones that targets enough of the feature in questions
- keep two versions of Arturo: one before the change (let's call this `arturo-1`) and another one after (let's say `arturo-2`) - this can easily be done via `./build.nims install --as:arturo-2`, etc
- run the test with Hyperfine (I have numerous scripts that automate this for specific scenarios, but none of them actually public in this repo). Nothing to fancy; all pretty much resulting in something like: `hyperfine "bin/arturo-1 testscript.art" "bin/arturo-2 testscript.art"`

beef331:
> we can see the impact of any given change on a rather varied codebase.
Sure, but it takes like 10+ minutes to run your benchmark suite and it'd be nice to see relatively quickly whether a change has benefical consequences locally without waiting 20+ minutes. For a change like this one, the benefits would appear on almost all Arturo code as far as I know.

drkameleon:
> Sure, but it takes like 10+ minutes to run your benchmark suite and it'd be nice to see relatively quickly whether a change has benefical consequences locally without waiting 20+ minutes.

What I usually do is a couple of runs of the tester: `arturo-1 tools/tester.art`. On my 2018 MacBook Pro, this takes like 25seconds total and it's more than a thorough test (it runs pretty much every examples from Rosetta Code + more).

Anyways, I'll have a look into it myself tomorrow morning (Spain time 😄 ) and I'll let you know about my findings as well. The core of your idea (reducing the memory needed for the storage of the Value objects) does make sense, so... let's see... 😉

beef331:
Yay I was not wrong about performance increase, I just missed another large type.

```
Benchmark 1: arturo-fast ./tests/core/integer_increment_1m.art
  Time (mean ± σ):     383.6 ms ±   4.8 ms    [User: 334.0 ms, System: 49.1 ms]
  Range (min … max):   373.9 ms … 399.4 ms    100 runs


Benchmark 1: arturo ./tests/core/integer_increment_1m.art
  Time (mean ± σ):     455.3 ms ±   3.2 ms    [User: 368.1 ms, System: 86.5 ms]
  Range (min … max):   450.2 ms … 468.6 ms    100 runs

```

Way outside of the margin of error.

drkameleon (approved):

drkameleon:
First of all, let me tell you: awesome job overall! 😉 

> Yay I was not wrong about performance increase, I just missed another large type.
> 
> ```
> Benchmark 1: arturo-fast ./tests/core/integer_increment_1m.art
>   Time (mean ± σ):     383.6 ms ±   4.8 ms    [User: 334.0 ms, System: 49.1 ms]
>   Range (min … max):   373.9 ms … 399.4 ms    100 runs
> 
> 
> Benchmark 1: arturo ./tests/core/integer_increment_1m.art
>   Time (mean ± σ):     455.3 ms ±   3.2 ms    [User: 368.1 ms, System: 86.5 ms]
>   Range (min … max):   450.2 ms … 468.6 ms    100 runs
> ```
> 
> Way outside of the margin of error.

I've just confirmed it myself. I ran several isolated tests, plus the whole thing (`tools/tester.art`) - master vs this branch - and I can see an increase (I would say something between 3-5%, not bad at all!). 🚀 

> Well now that we have a performance increase here, any direction on how you want to take these changes? Do you mind the templates for `funcType` or do you want to directly access fields through `funcType`?

Interesting question.

When I saw this `.action()()` syntax, I have to admit I was totally baffled and was about to tell you to change it. lol But... having had a look at what you're doing, I get it and I'd say I like it! 😄 

The same would go for `converter toDateTime...` - I don't know how but I had no idea this thing even existed! *Very* nice!

(That being said, as you may probably have guessed yourself already, I don't consider myself a Nim... guru by any means; the initial idea - and as a matter of fact, this is exactly what the project was written in - was something between C and D; Nim came only as an afterthought, after I tested it enough to see that the performance loss is practically non-existent, it came with good cross-platform support, plus it wasn't too difficult to get started with and would save me some of the typical - mostly C-related - headaches. However, I still think I must be doing several things wrong with respect to how I use Nim; hence I've spent hours and hours studying its generated code (and its source as well), but still...)

**Regarding the general direction:** I was thinking previously of investigating the Value objects (for things like the ones you've pointed out; the truth is I don't think I would have imagined `DateTime` as a possible culprit lol) - I guess I've been caught up a bit with different things, mostly the internals of the VM (which have given us an admittedly good boost).

The first one I was thinking in that aspect was, for some reason, the integers. The way integers are handled right now is a bit "off", code-wise. I mean... I want Arturo's `:integer` type to support both normal/64-bit `int`s and BigNums (whether these should be GMP-based by default, or if we could offer a more-flexible alternative is another question...). And this should happen in a totally "hidden" way for the user. However, I think it would be much more practical if they were actually treated as *one* type (e.g. `VInteger`) internally as well.

**Regarding changes:** the only one I would ask for - if you don't mind (I can obviously edit it myself) - would be to just rename the `FuncType` to `VFunction` (only to follow the example of all other "custom"/internal types, like `VBinary`, etc). 

Other than that, I'm totally with you! 😄 

> To the memory usage question. It's not overly accurate but Gnu's Time gives some insight in memory consumption. A comparison of the two is as follows.
> 
> ```
>         Command being timed: "arturo-fast ./tools/tester.art"
>         User time (seconds): 16.90
>         System time (seconds): 2.79
>         Percent of CPU this job got: 99%
>         Elapsed (wall clock) time (h:mm:ss or m:ss): 0:19.69
>         Average shared text size (kbytes): 0
>         Average unshared data size (kbytes): 0
>         Average stack size (kbytes): 0
>         Average total size (kbytes): 0
>         Maximum resident set size (kbytes): 118128
>         Average resident set size (kbytes): 0
>         Major (requiring I/O) page faults: 0
>         Minor (reclaiming a frame) page faults: 1543053
>         Voluntary context switches: 9386
>         Involuntary context switches: 1915
>         Swaps: 0
>         File system inputs: 0
>         File system outputs: 0
>         Socket messages sent: 0
>         Socket messages received: 0
>         Signals delivered: 0
>         Page size (bytes): 4096
>         Exit status: 1
> 
> 
>         Command being timed: "arturo ./tools/tester.art"
>         User time (seconds): 16.94
>         System time (seconds): 2.75
>         Percent of CPU this job got: 100%
>         Elapsed (wall clock) time (h:mm:ss or m:ss): 0:19.70
>         Average shared text size (kbytes): 0
>         Average unshared data size (kbytes): 0
>         Average stack size (kbytes): 0
>         Average total size (kbytes): 0
>         Maximum resident set size (kbytes): 120476
>         Average resident set size (kbytes): 0
>         Major (requiring I/O) page faults: 0
>         Minor (reclaiming a frame) page faults: 1546136
>         Voluntary context switches: 9407
>         Involuntary context switches: 1431
>         Swaps: 0
>         File system inputs: 0
>         File system outputs: 0
>         Socket messages sent: 0
>         Socket messages received: 0
>         Signals delivered: 0
>         Page size (bytes): 4096
>         Exit status: 1
> ```
> 
> For the integer test I did above:
> 
> * Master Branch: `Maximum resident set size (kbytes): 1091448`
> * This Branch: `Maximum resident set size (kbytes): 617880`

I'm mostly using macOS's `time` command, which AFAIK differs a bit from its Linux equivalent. So, I usually do something like `/usr/bin/time -lp ...` which gives enough information about memory usage and perhaps even more:

Screenshot 2022-11-07 at 11 40 59

It works rather ok; only thing is it doesn't work with Hyperfine, so getting this type of info would mean running the tests twice - and then it would end up being too much. I don't know... (one idea I've had - especially for the benchmark suite - is to have hyperfine run all tests as many times as it wishes, and run a `time` tool once, to get the memory consumption as a "sample"...)

-----

As a totally-irrelevant sidenote, here's an interesting issue you might want to have a look into: https://github.com/arturo-lang/arturo/issues/792 (this thing has been bugging me too much, but I don't know where to find the time to do all of it myself. so any help - regarding this one, or anything else for that matter - would be greatly appreciated! 😄 )

beef331:
It might be worth reducing the size smaller and using a pointer for all rarer types. Version for instance is 5 pointer wide. Doing that will likely reduce performance in cases that use these objects due to pointer indirection, but for other common types increase performance and reduce memory consumption further. Hopefully the warning and hint I added suffice and catch this in the future.

drkameleon:
> It might be worth reducing the size smaller and using a pointer for all rarer types. Version for instance is 5 pointer wide. Doing that will likely reduce performance in cases that use these objects due to pointer indirection, but for other common types increase performance and reduce memory consumption further. Hopefully the warning and hint I added suffice and catch this in the future.

Just checked it out. I absolutely *love* the hint idea (so that I don't have to keep counting every single time something changes). Simply great!

It generally looks good to me. So... I think we are ready to merge! 🚀 

-----

Regarding the Value object's size, I think `Version` values would be the next candidate to squeeze (ofc, they are not common by any means, but obviously they do contribute to a larger overall size). Blocks could be the next one? (right now, I mean as of the later PR I have in the works, they have a Value, a seq[Value] and 2 *bool* fields).

In any case, thanks a lot! Feel free to pick any pending issue or ask me if you need... inspiration. You're more than welcome! 😉

drkameleon (approved):

beef331:
With Orc it's worth noting `seq` and `string` store the length on the stack so they're 16 bytes wide. So it'll be a fun game of balances.
`info` is a string, but could it just be an index to some collection on the VM?
All your `bool`s could be turned into a `ValueFlag` and then you could do like `boolVal in value.flags` or `dirty in value.flags` to reduce size(remember depending on alignment a `bool` can take up to 8 bytes). Reducing data type sizes obviously is not a silver bullet but it can go a long way into performance.

Issue #795, opened 2022-11-07
Shrink 'Value' type further


beef331:
Presently the big memory usage is the `ValueDict` as `OrderedTable` is 5 pointer widths wide, this creates a larger object.

PR #796, opened 2022-11-08
WIP: Reduced footprint of `Value` even further

beef331:
# Description

Following more investigations I looked at making `Value` even smaller, but this pass might result in slower performance for objects/dictionaries so benchmarks really need to be looked at. Presently on a 64bit machine Value is now 32bytes instead of 56.
```
Benchmark 1: arturo ./integer_increment_inc_inplace_1m.art
  Time (mean ± σ):     249.9 ms ±   8.0 ms    [User: 215.9 ms, System: 33.6 ms]
  Range (min … max):   242.5 ms … 269.4 ms    11 runs

Benchmark 1: arturo-fast ./integer_increment_inc_inplace_1m.art
  Time (mean ± σ):     239.6 ms ±   3.8 ms    [User: 214.7 ms, System: 24.6 ms]
  Range (min … max):   236.2 ms … 250.3 ms    12 runs

Benchmark 1: arturo ./dictionary_lookup_1m.art
  Time (mean ± σ):     456.4 ms ±   7.5 ms    [User: 395.5 ms, System: 58.5 ms]
  Range (min … max):   449.8 ms … 472.2 ms    10 runs

Benchmark 1: arturo-fast ./dictionary_lookup_1m.art
  Time (mean ± σ):     437.9 ms ±   2.6 ms    [User: 395.2 ms, System: 42.1 ms]
  Range (min … max):   435.8 ms … 442.7 ms    10 runs
```

Some memory looks:
```
/usr/local/bin/time -v arturo ./array_access_1m.art
Maximum resident set size (kbytes): 634336

/usr/local/bin/time -v arturo-fast ./array_access_1m.art
Maximum resident set size (kbytes): 476692

/usr/local/bin/time -v arturo ./integer_increment_1m.art
Maximum resident set size (kbytes): 617996

/usr/local/bin/time -v arturo-fast ./integer_increment_1m.art
Maximum resident set size (kbytes): 460448
```

Addresses some parts of https://github.com/arturo-lang/arturo/issues/795

Fixes # (issue/s)

## Type of change

- [x] Code cleanup

### DIFF
--- a/src/library/Converters.nim
+++ b/src/library/Converters.nim
@@ -1064,21 +1064,21 @@ proc defineSymbols*() =
                                 else:
                                     options[k] = (vspec, "")
 
-                        ret.attrs() = options
+                        ret.attrs = options
 
                     if (let returnsData = y.data.d.getOrDefault("returns", nil); not returnsData.isNil):
                         if returnsData.kind==Type:
-                            ret.returns() = {returnsData.t}
+                            ret.returns = {returnsData.t}
                         else:
                             var returns: ValueSpec
                             for tp in returnsData.a:
                                 returns.incl(tp.t)
-                            ret.returns() = returns
+                            ret.returns = returns
 
                     if (let exampleData = y.data.d.getOrDefault("example", nil); not exampleData.isNil):
-                        ret.example() = exampleData.s
+                        ret.example = exampleData.s
     
-            ret.args() = argTypes
+            ret.args = argTypes
             
             push(ret)
 
--- a/src/vm/exec.nim
+++ b/src/vm/exec.nim
@@ -274,11 +274,11 @@ proc execFunction*(fun: Value, fid: Hash) =
         # pop argument and set it
         SetSym(arg.s, move stack.pop())
 
-    if fun.bcode().isNil:
-        fun.bcode() = newBytecode(doEval(fun.main))
+    if fun.bcode.isNil:
+        fun.bcode = newBytecode(doEval(fun.main))
 
     try:
-        ExecLoop(fun.bcode().trans.constants, bcode(fun).trans.instructions)
+        ExecLoop(fun.bcode.trans.constants, bcode(fun).trans.instructions)
 
     except ReturnTriggered:
         discard
--- a/src/vm/exec.nim
+++ b/src/vm/exec.nim
@@ -239,7 +239,7 @@ proc execFunction*(fun: Value, fid: Hash) =
     ##   pretty much like `execLeakless`
 
     var memoizedParams: Value = nil
-    var savedSyms: ValueDict
+    var savedSyms: ValueStackDict
 
     var savedArities = Arities
     let argsL = len(fun.params.a)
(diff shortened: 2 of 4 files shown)


beef331:
Odd 418 errors fail on my machine so seems like there might be some UB with the `sink ValueStackDict` converter.

drkameleon:
I read your open issue and suggestions last night and started thinking about a possible implementation and now I see you've already done work on it - awesome! 😄 

The Version value going to the heap is definitely an obvious one - personally, I would also move to a separate module (it doesn't depend on Value, so it's "pure" enough); that is: under `vm/values/custom`, like the other ones, along with its corresponding `==`, `>`, overload, etc. (from vm/values/comparison).

Your implementation for flags is pretty much what I had in mind. Very neat. The use of properties for Function value fields also looks nicer. 

Also, let me comment on the `makeConst` - I *love* it. It definitely looks cleaner to the eye and far less repetitive! Good job! 😉 

Now... for the ValueDict's... That's a tricky one. Perhaps, it would qualify as one of the aspects I would least dare to touch but let's see...

I'll pull the branch and test it locally to see what is going on (although, to be perfectly honest, my instict said from second one when I read that there were failing tests that scoping issues would be the first thing to expect - and there they are... lol).

Let's see...

drkameleon:
And another comment:

> Presently on a 64bit machine Value is now 32bytes instead of 56.

Yesterday, when I merged master into [the branch I'm working on locally](https://github.com/arturo-lang/arturo/pull/787), the size of the Value objects was 72 (!?). And I think the only difference is that I have added another `bool` in the Block values.

That on a 2.3 GHz Quad-Core Intel Core i5, 2018 MacBook Pro, running macOS Monterey (12.4).

drkameleon (commented):
I've made a couple of changes to make the CI actions succeed. 💥 

Generally, it looks fine. The only things that I'm skeptical about is the conversion of `ValueDict` to a *ref* - as you yourself mentioned. We may be gaining something in terms of object size & memory, but the results so far don't convince me this is really faster... 

(I've run some 5-6 tests so far, comparing this branch - with the fixes - and master and it's a pretty consistent 26 vs 28s, with the faster being the previous one; judging by the core role of `ValueDict`s throughout the codebase, I cannot think of anything else...)

That being said, I think I'll have to test it a tiny bit more thoroughly to make sure.

That, solely regarding `ValueDict`; the rest looks more than safe. 😉

beef331:
As I suspected, this was not that beneficial, so there is no point moving `Version` to ref as `ValueDict` is what's taking a lot of memory up. Though it's odd that your version reports 48 bytes, mine is 32.

drkameleon:
> I ran a few tests from the rosetta examples and some are on par but some ar slower. So as I suspected, this was not that beneficial

I'll measure it more extensively tomorrow morning to see what's going on...

> , so there is no point moving `Version` to ref as `ValueDict` is what's taking a lot of memory up. 

`Version` values being *ref* doesn't bother me at all. As a matter of fact I had already though of doing it, although more for keeping it all tidy.

> Though it's odd that your version reports 48 bytes, mine is 32.

It is indeed. 🤔  Could it be the compiler?! 

In my case:

```
Apple clang version 13.1.6 (clang-1316.0.21.2.5)
Target: x86_64-apple-darwin21.5.0
Thread model: posix
```

beef331:
Oh whoops I was not passing `--mm:orc` for my compiling, so it was smaller.

beef331:
Closing this for a less invasive PR.

PR #797, opened 2022-11-08
Introduce "leakless" functions

drkameleon:
# Description

Right now, a function call creates a whole new scope, which is something that doesn't happen with e.g. iterator blocks. 

In an iterator block (let's say `loop 1..10 'x [ print x ]`):

- The `x` symbol will be protected
- Every other symbol declared inside the block will propagate to the external scope

So... what if - for a specific subset of functions (not all of them) - we don't need a whole scope, but perhaps a function is only doing something with the passed argument(s) and returns a value (or not)?

In that case, we could introduce the concept of "leakless" or inline(?) functions.

This can be declared either explicitly (by some option? e.g. `.inline`) or implicitly (if - when generating the Function value - we are able to verify that the function is safe to consider a "leakless" one).

Once this is established, the functions marked as such will be executed as "leakless" blocks (pretty much like an iterator block - only without any values being propagated), and thus potentially much faster.

## Type of change

- [x] Code cleanup
- [x] Performance enhancement


drkameleon:
This PR brings another 3-5% speedup along with an extreme boost in specific scenarios. 💥 

Let's take the following [recursive fibonacci code](https://github.com/arturo-lang/arturo/blob/master/examples/rosetta/fibonacci%20sequence%20-%20recursive.art) for example:

```red
fib: $[x][
	if? x<2 [1]
	else [(fib x-1) + (fib x-2)]
]

loop 1..25 [x][
	print ["Fibonacci of" x "=" fib x]
]
```

And these are the results from Hyperfine (comparing Arturo *before* and *after* this PR):

```
Benchmark 1: bin/arturo-before examples/rosetta/fibonacci\ sequence\ -\ recursive.art
  Time (mean ± σ):      8.373 s ±  0.197 s    [User: 8.121 s, System: 0.081 s]
  Range (min … max):    8.062 s …  8.595 s    10 runs
 
Benchmark 2: bin/arturo-after examples/rosetta/fibonacci\ sequence\ -\ recursive.art
  Time (mean ± σ):     351.7 ms ±  23.3 ms    [User: 326.3 ms, System: 10.3 ms]
  Range (min … max):   328.1 ms … 411.2 ms    10 runs
 
Summary
  'bin/arturo-after examples/rosetta/fibonacci\ sequence\ -\ recursive.art' ran
   23.81 ± 1.68 times faster than 'bin/arturo-before examples/rosetta/fibonacci\ sequence\ -\ recursive.art'
```

I guess I didn't even need a benchmark to see that the previous version was... desperately slow. But what is more than certain is that the performance boost here is tremendous! (We're talking about ~24x faster! 😄 )

-----

So... without further ado....
ready to merge! 🚀

PR #798, opened 2022-11-09
Re-work arity checking and handling

drkameleon:
# Description

Right now, we keep 2 tables at any given time: 

- one for the Symbols (`Syms`) 
- another one for the Arities 

...which we copy & move around relentlessly until we help performance ... drop sufficiently (lol).

Given the - many - recent changes to scoping and how the evaluator works (practically, evaluating much less often and producing more optimized code), I think we could improve it a lot. Even, possibly eliminate the need for an Arities table altogether, which would give us yet another performance boost. 🚀 

So,... let's go for it! 😄 

## Type of change

- [x] Code cleanup
- [x] Performance enhancement


drkameleon:
Far easier than initially thought (again!) and quite smooth in general, with great results (another 8-10% boost, at least 💥 ).

Ready to merge! 🚀

PR #800, opened 2022-11-09
Changed `Value`  to use flags and `VVersion`

beef331:
# Description
Moved Version to `VVersion`.
Moved bools to flags, this presently does not change much, but it'll always be better to have flags  then individual bools if this grows. Also removed spurious templates and moved to properties.

I know this is a lot for a single PR, but sorry :stuck_out_tongue: 

## Type of change

- [x] Code cleanup


drkameleon:
Awesome job, again! 😄 

I think the approach of going for a less pervasive change is a good one.

I've been benchmarking it thoroughly and I have some doubts about one thing, but let me go through it point by point:

- The `makeConst` idea; I *love* it - much cleaner, as I've said before
- Packing `Version` value as a ref object makes sense; it's not a commonly used Value type, as you can already imagine, but it makes the whole thing look more organized in any case
- Function value accessors behaving like properties also looks fine

Now, I have some issues with the Flags idea... only in respect to one thing: Logical values.

I'm not sure if it has to do with how Logical values are currently being used throughout the code (I think it does), but I've noticed some performance drop in various cases/benchmarks and I believe the culprit must be this. Using a proc call - even though it might be an inline one - with if/else inside, while we previously had just a field access, must have some impact.

For example,  right now, taken from the implementation of `if` (which has already changed as a matter of fact: what the latest PR introduces is proper inlining, as bytecode, of many of these constructs; up to now, `if` was just another function - after the PR in question, `if` will still be a function, but if the evaluator thinks it's a good idea, it will be converted to bytecode directly with the proper jumps & branching in-place 😉 ):

```nim
let condition = not (x.kind==Null or (x.kind==Logical and x.b==False))
    if condition: 
        execUnscoped(y)
```

This, with `.b` being a mere field, makes some sense. With `.b` calling a function and not actually being a field looks like a total and absolute overkill... 

For example, for this very piece of code, what I would think is:
- have `isFalse` as another possible flag in the set
- since only Logical values can have an `isFalse` flag set, we can omit `x.kind==Logical`
- we want to test if this value is false, so a `isFalse in x.flags` (or a template doing the same thing) would suffice

And in that case, not only it wouldn't cause a performance drop, but I think it could be the opposite.

But still, I think it has to do with the flags being used for boolean/Logical values (which are ultra-common and used in various places) + my implementation of various things, based on how it was up to now...

Let me experiment a bit and I'll let you know about my conclusions. 😄

src/vm/values/types.nim
```diff
@@ -234,13 +240,63 @@ type
...
-    assert val.kind == Function
-    val.funcType.name
+proc readonly*(val: Value): bool {.inline.} = isReadOnly in val.flags
+proc `readonly=`*(val: Value, newVal: bool) {.inline.} = val.flags[isReadOnly] = newVal
+
+proc dirty*(val: Value): bool {.inline.} = isDirty in val.flags
+proc `dirty=`*(val: Value, newVal: bool) {.inline.} = val.flags[isDirty] = newVal
+
+proc dynamic*(val: Value): bool {.inline.} = isDynamic in val.flags
+proc `dynamic=`*(val: Value, newVal: bool) {.inline.} = val.flags[isDynamic] = newVal
```
drkameleon, line 253:
```suggestion
template `dynamic=`*(val: Value, newVal: bool) = val.flags[isDynamic] = newVal
```

src/vm/values/types.nim
```diff
@@ -234,13 +240,63 @@ type
...
+    else:
+        val.flags.excl {isMaybe, isTrue}
+
+
+template makeVersionAccessor(name: untyped) =
+    proc name*(val: Value): typeof(val.version.name) {.inline.} =
+        assert val.kind == Version
+        val.version.name
+
+    proc `name=`*(val: Value, newVal: typeof(val.version.name)) {.inline.} =
```
drkameleon, line 283:
```suggestion
    template `name=`*(val: Value, newVal: typeof(val.version.name)) =
```

src/vm/values/types.nim
```diff
@@ -234,13 +240,63 @@ type
...
+        assert val.kind == Version
+        val.version.name = newVal
+
+makeVersionAccessor(major)
+makeVersionAccessor(minor)
+makeVersionAccessor(patch)
+makeVersionAccessor(extra)
+
+template makeFuncAccessor(name: untyped) =
+    proc name*(val: Value): typeof(val.funcType.name) {.inline.} =
```
drkameleon, line 293:
```suggestion
    template name*(val: Value): typeof(val.funcType.name) =
```

src/vm/values/types.nim
```diff
@@ -234,13 +240,63 @@ type
...
+        assert val.kind == Function
+        val.funcType.name
+
+    proc `name=`*(val: Value, newVal: typeof(val.funcType.name)) {.inline.} =
```
drkameleon, line 297:
```suggestion
    template `name=`*(val: Value, newVal: typeof(val.funcType.name)) =
```

src/vm/values/types.nim
```diff
@@ -234,13 +240,63 @@ type
...
+
+proc dirty*(val: Value): bool {.inline.} = isDirty in val.flags
+proc `dirty=`*(val: Value, newVal: bool) {.inline.} = val.flags[isDirty] = newVal
+
+proc dynamic*(val: Value): bool {.inline.} = isDynamic in val.flags
+proc `dynamic=`*(val: Value, newVal: bool) {.inline.} = val.flags[isDynamic] = newVal
+
+proc b*(val: Value): VLogical {.inline.} =
+    assert val.kind == Logical
+    assert (val.flags * {isMaybe, isTrue}).len <= 1 # Ensure we do not have {isMaybe, isTrue}
```
drkameleon, line 257:
```suggestion
```

Asserts are fine; only thing is `-d:danger` (which is what Arturo is compiled with) totally ignores them.

src/vm/values/types.nim
```diff
@@ -234,13 +240,63 @@ type
...
+    if val.flags * {isMaybe} == {isMaybe}:
+        Maybe
+    elif val.flags * {isTrue} == {isTrue}:
+        True
+    else:
+        False
```
drkameleon, line 263:
```suggestion
    if isTrue in val.flags: True
    elif isFalse in val.flags: False
    else: Maybe
```

Could we add another `isFalse` flag? I think it would many things simpler (since a) we know by definition that only Logical values can have them, b) If a Value is a Logical one, one of the three *will* exist no matter what). 

Also it would help create helpers like `template IsFalse*(v: Value):untyped = isFalse in v.flags` for fast checks (to replace things like `x.b == True` which look like a total overkill this way, but are pretty much everywhere...)

src/vm/values/types.nim
```diff
@@ -147,38 +147,46 @@ type
...
             of BuiltinFunction:
                 action*     : BuiltinAction
+    VVersion* = ref object
+        major*   : int
+        minor*   : int
+        patch*   : int
+        extra*   : string
+
+    ValueFlag* = enum
+      isDirty, isDynamic, isReadOnly, isTrue, isMaybe
```
drkameleon, line 157:
```suggestion
      isDirty, isDynamic, isReadOnly, isTrue, isFalse, isMaybe
```

Let's add another `isFalse` flag (see below 😄 )

src/vm/values/types.nim
```diff
@@ -234,13 +240,63 @@ type
...
+    assert val.kind == Logical
+    case newVal
+    of Maybe:
+        val.flags.excl isTrue
+        val.flags.incl isMaybe
+    of True:
+        val.flags.excl isMaybe
+        val.flags.incl isTrue
+    else:
+        val.flags.excl {isMaybe, isTrue}
```
drkameleon, line 275:
```suggestion
    of True:
        val.flags.excl {isFalse, isMaybe}
        val.flags.incl isTrue
    of False:
        val.flags.excl {isTrue, isMaybe}
        val.flags.incl isFalse
    else:
        val.flags.excl {isTrue, isFalse}
        val.flags.incl isMaybe
```

Possible implementation, after having added an extra `isFalse` flag

src/vm/values/value.nim
```diff
@@ -528,11 +537,21 @@ func newBytecode*(t: sink Translation): Value {.inline, enforceNoRaises.} =
 
 func newInline*(a: sink ValueArray = @[], dirty = false): Value {.inline, enforceNoRaises.} =
     ## create Inline value from ValueArray
-    Value(kind: Inline, a: a, dirty: dirty)
+    let flags =
+      if dirty:
+          {isDirty}
+      else:
+          {}
```
drkameleon, line 544:
```suggestion
    let flags =
        if dirty:
            {isDirty}
        else:
            {}
```

Adjusted indentation to 2 tabs

src/vm/values/value.nim
```diff
@@ -528,11 +537,21 @@ func newBytecode*(t: sink Translation): Value {.inline, enforceNoRaises.} =
...
+    Value(kind: Inline, a: a, flags: flags)
 
 func newBlock*(a: sink ValueArray = @[], data: sink Value = nil, dirty = false): Value {.inline, enforceNoRaises.} =
     ## create Block value from ValueArray
-    Value(kind: Block, a: a, data: data, dirty: dirty)
+    let flags =
+      if dirty:
+          {isDirty}
+      else:
+          {}
```
drkameleon, line 553:
```suggestion
    let flags =
        if dirty:
            {isDirty}
        else:
            {}
```

Again, indentation adjustment.

drkameleon (commented):
So... after 3 trillion experiments (I've been benchmarking one version against the other for at least the past 5 hours 🥴, literally)... my main conclusions - most of which you'll find in the comments above (If it looks messy, my apologies - as you can see for yourself I'm not a master at reviewing PRs lol):

- As incredible as it sounds, making the accessors inline `proc`s does have an impact. I have tested it thoroughly, again and again, and despite what I have read over and over, and despite what *I* would expect: Timings are different (slightly slower) + the final binary is like 30-40KB more in size. So, let's move them back to templates.

- As I said, the `VVersion`, etc looks fine

- Regarding the flags system, I generally agree (as you can see, we already have another necessary `bool` - `isDynamic` - so that went into the flags as well). Right now, the performance is practically identical to the master branch (perhaps a tiny bit slower). I think with the necessary changes + changing all my `.b` access points to actually check if a value is true or false (with something like the template I describe in one of the comments), the performance could improve even more...

Let's see! 😄

drkameleon:
As a side idea (not that it's that important), we could generate our accessors in an even more uniform way, with something like:

```nim
macro createAccessors*(root: untyped, fields: varargs[untyped]) =
    result = newStmtList()
    for field in fields:
        result.add quote do:
            template `field`*(val: Value): typeof(val.`root`.`field`) =
                val.`root`.`field`

            template `field=`*(val: Value, value: typeof(val.`root`.`field`)) =
                val.`root`.`field` = value
```

And use it like:

```nim
createAccessors(funcType,args,attrs,arity,returns,example,fnKind,params,main,imports,exports,memoize,inline,bcode,action)
```

(It seems like I just wanted to play with macros - which I did lol - but the more things like this we have, the more a "shortcut" would help to keep it clean 😉 )

src/vm/values/types.nim
```diff
@@ -234,28 +240,67 @@ type
...
+
+template isFalse*(val: Value): bool = isFalse in val.flags
+template isMaybe*(val: Value): bool = isMaybe in val.flags
+template isTrue*(val: Value): bool = isTrue in val.flags
+
+
+template makeAccessor(field, subfield: untyped) =
+    template subfield*(val: Value): typeof(val.field.subfield) =
+        let myVal = val.field.subfield
+        myVal
```
drkameleon, line 278: The changes are all looking good. Good job!

I have only one issue:

Is there some specific reason this ^ is done this way?

I mean... I generally tend to avoid every possible `let` or `var` (since it usually leads to unnecessary copying).

In that case, given that these fields are used quite a lot (e.g. on function calls) and some of the fields are arrays (`seq`s), I'm looking into the generated C code and I'm seeing quite a few `eqcopy`s for this `myVal`.

Couldn't we just "alias" subfield with `val.field.subfield`? Is there some issue with that approach?

beef331: Whoops that was me fighting around the issue when `b` is a template.

drkameleon: `.b` that is trying to get a Logical value's value... doesn't (or won't better said) won't matter too much.

The way you've cleaned up the code has given me a great idea:

As I've said, in many many instances, I check for things like: `if x.kind==Logical and x.b == True`, etc. Given the way `Logical` values are defined right now, this whole construct will be transformed to a simple `isTrue(x)`. And obviously, we won't really have to bother what `.b` is... 😉

drkameleon:
Great job with this PR 😄 - I'm not sure there is such an enormous gain performance-wise; I would say it say it generally fares pretty much the same (perhaps something like 0.5% better), but still: I think in the long run, especially your Flags idea, its will bear its fruits. 

I just adjusted a couple of things (mostly how this `x.kind==Logical and x.b==True` type of things were written, to be in line with our new set up)

And... ready to merge! 🚀 

PS. Thanks again - keep them coming! 😉

PR #810, opened 2022-11-13
Made value's info a ref string to further reduce value's size

beef331:
# Description

Simply just made `info: string` into `infoRef: ref string` and expose the string using `info` properties. This yet again gains speed for the VM

Fixes # (issue/s)

## Type of change

- [x] Code cleanup


drkameleon (approved):
That's a nice one. Nothing to comment on here! Even if `.info` is eliminated at some point, it's still a step in the right direction! Thanks! 😉 

And... ready to merge! 🚀

PR #811, opened 2022-11-14
Convert SymTable to an unordered Table & misc. related optimizations

drkameleon:
# Description

_Inspired by the discussion @ https://discord.com/channels/765519132186640445/829324913097048065/1040204652966330368_

-----

Right now, there is some heavy use of tables throughout the code: mainly, for the core Symbol Table and the Dictionary values, all of them being declared as [`OrderedTable`](https://nim-lang.org/docs/tables.html#OrderedTable). 

Given the size of these (compared to a normal table - given the extra `seq` needed to keep track of the order) + the overhead of managing them, eliminating them (or having a more efficient implementation) could be an interesting idea.

However, an even easier one - also made possible because of the recent changes in how Eval/Exec work with respect to handling blocks and/or Dictionary blocks - is to simply re-define *where* we need them. And we most definitely don't need them as often as they appear.

**In a few words:**

- In order for Dictionary values to maintain their current behavior - which follows the general rule and I personally *love* - we'll keep using ordered tables (that is: `ValueDict` objects)
- But: we don't need ordering for the Symbol Table; it pretty much doesn't matter anymore since (a) the 2 concepts - Symbol Tables and Dictionaries - are totally unrelated as of recently, (b) the cost of managing an ordered table for something as common as storing/reinstating/etc symbol tables is far higher.

So... let's give it a try! 😄 (and I totally have a hunch this will give us another considerable performance boost)

## Type of change

- [x] Code cleanup


drkameleon:
Another ~15-20% ⬆️  up in performance and... 

ready to merge! 🚀

PR #825, opened 2022-11-24
Convert ValueDict objects to references

drkameleon:
# Description

Inspired by @beef331's idea: https://github.com/arturo-lang/arturo/pull/796

Basically, the idea was fine. The problem is that - at the time of the PR - the ValueDict objects were also used as the main symbol table (which was changed later on), and thus the change created quite a few issues.

Fixes: #795 

## Type of change

- [x] Code cleanup
- [x] Performance enhancement


drkameleon:
After, the performance gain speed-wise is roughly 2-3%.

Memory-wise, and obviously depending on the test in question, the drop in memory consumptiyon might range from 2-10%.

All in all, the change is totally sound and makes perfect sense.

Thanks @beef331 for the suggestion! 😉 

-----

Ready to merge! 🚀

drkameleon:
Ready to merge! 🚀

PR #863, opened 2022-12-12
Adds FFI example to call function

RickBarretto:
# Description

Adds missing example of how to properly use FFI in Arturo.
>Note: It'll cover both `.expect` and `.external` 

The current documentation does not cover this feature, and I had to search into Rosetta examples to discover how to use it:
[call a foreign-language function.c](https://github.com/arturo-lang/arturo/blob/master/examples/rosetta/call%20a%20foreign-language%20function.c)
[call a foreign-language function.art](https://github.com/arturo-lang/arturo/blob/master/examples/rosetta/call%20a%20foreign-language%20function.art)

## Type of change

### DIFF
--- a/src/library/Core.nim
+++ b/src/library/Core.nim
@@ -129,23 +129,22 @@ proc defineSymbols*() =
             ..........
             call $[x][x+2] [5]            ; 7
             ..........
-            // Let's say you want to call external functions:
-            //
-            // mylib.c
-            // compile with:
-            // clang -c -w mylib.c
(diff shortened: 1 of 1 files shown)


drkameleon:
I made a tiny fix (as mentioned above, to convert the C code in the example in an Arturo comment).

Other than that, great idea!!

I think I had forgotten about since I considered this feature a tiny bit... not-very-mature, but since it's there (and it doesn't appear to work that bad), why not have it in the documentation as you said...

Thanks! 😉 

------

Ready to merge! 🚀

PR #866, opened 2022-12-14
Implement global configuration & add new persistent Store values

drkameleon:
# Description

There are different functions (e.g. **[`mail`](https://arturo-lang.io/documentation/library/net/mail/)**) that take a number of parameters that may - or may not - need not be changed.

For example, let's say I'm using Arturo to send out mails. Basically, the most essential parts of this idea is: set the message & the recipient (and most likely the title).

Of course, I'd also need to specify the SMTP server and username/password for this SMTP server, and so on.

But what are the chances I'm going to need to change all this for every single time I want to send an e-mail?

**So, the idea is:**

- Have the option to set them, when we need to
- Pull them from some global configuration file
- Manually passed options would override the ones from the configuration file

(Of course, the idea could also include many other things, not only related to commands like `mail`... even about how Arturo behaves...)

The idea could also be extended to actually be able to get/set configuration option through a command.

But since getting/setting is already possible for Dictionary values (e.g. `mydict/someKey` or `mydict/someKey: someVal`), why not use that?

Or... over-extending the idea, why not have a new type (`:configuration`?) that behaves like a Dictionary, but actually stores the whole thing on the disk automatically?

Hmm, hmm... Let's see! 🚀 

-------

The final implementation is what one would call a persistent, hash/dictionary-based "store", e.g. Python's [`shelve`](https://docs.python.org/3/library/shelve.html) or Ruby's [`pstore`](https://ruby-doc.org/stdlib-2.5.3/libdoc/pstore/rdoc/PStore.html).

Arturo's `:store` values can be auto-saveable (or saved just before the app terminates), support native stores (in Arturo itself), JSON or SQlite (for now!) and the point is they have to be as flexible to use as normal Arturo dictionaries. Only they'll... persist 😉 

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #874, opened 2022-12-15
[Collections/remove] add `.instance` option

### LINKED ISSUE #870: [Collections/remove] should we add a `.value` option?
[Collections/remove] should we add a `.value` option?

For example: ```

debug remove [1 [6 2] 5 3 [6 2] 4 5 6] [6 2]

```

This removes "6" & "2", but what if we want to actually remove all [6 2] *instances*?

https://github.com/arturo-lang/arturo/blob/1136d74ff6685ec5b16c64021b3cf1a0dfc45fa8/src/library/Collections.nim#L1129

```text

    # TODO(Collections/remove) is `.index` broken?
    #  Example: `remove.index 3 'a, debug a`
    #  labels: library, bug

    # TODO(Collections/remove) should we add a `.value` option?
    #  For example: ```
    #   debug remove [1 [6 2] 5 3 [6 2] 4 5 6] [6 2]
    #  ```
    #  This removes "6" & "2", but what if we want to actually remove all [6 2] *instances*?
    builtin "remove",
        alias       = doubleminus,
        rule        = InfixPrecedence,
```

f4f17e747b843c26e9c7be40715e2c406a216145

RickBarretto:
# Description

Adds `.instance` attribute to `remove` function from Collections.

Fixes #870 , Note, I named it as `instance` instead of `value` to be a bit more semantic.

Running on my local build:

image
remove instance once

## Behaviors to be added
- [x] default
- [x] once
- [x] ~multiple instances?~ (closed discussion)

> ~Note: multiple instances should be pushed into another PR, using `instances` instead of `instance`.~

## Type of change

- [x] Add unit tests
- [x] Add usage example

## Requires
- [ ] Documentation update
- [ ] Build update
- [x] Unit test output update


RickBarretto:
Note: Remove first occurrence of an instance behavior was already implemented using just `remove.once`:
image

drkameleon:
Good job!

Regarding not using `.value` as the name of this type of behavior, I totally agree - it does not sound very descriptive. (As for `.instance`, I don't particularly adore it either, but it's surely more illustrative + I already have in mind going through some of the options and cleaning them up sort-of, so... no problem at all, let's leave it as-is for now!)

Regarding the different uses, I'm a bit confused with `.once` and the example you gave above. I'll have to re-test it myself and see.

As for attributes/options in general, the "rule" should be:

- as few as possible (no reason to create too many, unless they absolutely make sense)
- as descriptive as possible
- as short as possible
- if they make sense in combinations, then they are to be used this way (for example: `remove.once` vs `remove.instance.once`)
- if there is an option used throughout the library in a specific context, then we'd better use this one for all similar cases (and not create different names for roughly the same thing, since this will end up being too confusing - I'm giving you an example: if there was a case where a different built-in function does something along the lines of `.instance` and uses a different name for that option, I would definitely contemplate using the same name. But I'm not very sure there is one like this...)

drkameleon:
OK... so let's leave the unit-tests part for now and let's stick to what `remove` does/should-do.

Basically, given your `.instance` implementation and what I consider a bug for `.once`, this is what I think the logic should be:

Let's say we have an array...

`arr: [1 2 [1 2] 3 4 1 2 [1 2] 3 4]`

| | with simple value<br>(e.g. `remove arr 2`) | with Block<br>(e.g. `remove arr [1 2]`) | 
|---|---|---|
| no option |  ✅ remove all instances of value within collection:<br>`[1 [1 2] 3 4 1 [1 2] 3 4]`| ✅ replace all instances of all Block values within collection:<br>`[[1 2] 3 4 [1 2] 3 4]` |
| `.instance` | ✅ same as above - no effect:<br>`[1 [1 2] 3 4 1 [1 2] 3 4]` | ✅ remove all instances of *given* Block - the block itself - value within collection:<br>`[1 2 3 4 1 2 3 4]` |
| `.once`  | ✅ remove only first occurence of value:<br>`[1 [1 2] 3 4 1 2 [1 2] 3 4]` | ❌ remove only first of occurence of any Block value:<br>`[2 [1 2] 3 4 1 2 [1 2] 3 4]`<br>(it now returns: `[1 2 3 4 1 2 [1 2] 3 4]`) |
| `.instance.once` | ✅ same as above - no effect:<br>`[1 [1 2] 3 4 1 2 [1 2] 3 4]` | ❌ remove only the first occurence of *given* Block - the block itself - value within collection:<br>`[1 2 3 4 1 2 [1 2] 3 4]`<br>(it now returns: `[1 2 3 4 1 2 3 4]`) |

As you can see, there is no real need for `.instances` or any other complicated options - as I said, if we don't *really* need more options, we may use the existing ones efficiently. Plus, the above with `.once` is obviously a bug.

If you can make the two cases above (with the ❌) work, I would consider it more than ultra-complete! 😉

RickBarretto:
Alright, a lot of things to discuss here. So, I will put into topics, to be better.

## Renaming?

> `.instance`, I don't particularly adore it either, but it's surely more illustrative 

If you want to rename, I can. 😉

## Instance once Instance once

About the [explanation](https://github.com/arturo-lang/arturo/pull/874#issuecomment-1354693810), I got your point and agree with it.
When I was doing `.instance` attribute, my plan was to remove the every instance of B that exists in A. Alright, everything is running well. eg.: `remove.instance block instances`
But, then I got the idea to join `.instance` and `.once` to just remove the first instance that exists in A. eg.: `remove.instance.once block instances`

### The Once bug

When I did the above, I've noticed that just `.once` has the same behavior what would be `.instance.once`. LOOOL!
As you said, once must remove just the first element (not instance).

Also, another open discussion here, do you think that once should remove once per item from B or just the first item from B?

Your example was: 
```
remove.once  [1 2 [1 2] 3 4 1 2 [1 2] 3 4]  [1 2]; => [2 [1 2] 3 4 1 2 [1 2] 3 4]
; just removed 1, that is the first element of B
```

But, IMHO, it would behavior better with:
```
remove.once  [1 2 [1 2] 3 4 1 2 [1 2] 3 4]  [1 2]; => [[1 2] 3 4 1 2 [1 2] 3 4]
; notice that it removed both 1 and 2, but just the first occurrence.
```

## Instances? (closed topic)

~The idea of `.instances` is to check a block that contains blocks to remove. It would be like the original remove (that removes item by item of A from B), but with instances.~

> When I was writing it, I discovered that it's even possible now. Here is how:
> 
> image

So, topic is closed! 😉 

## How to fix it

Say what do you think, and if I must open new PR to do it, or just put everything in this. (I prefer open new, IMO)
I think the better approach is create a PR to fix `once` behavior, and then merge it when `.instance.once` is done!

drkameleon:
> once should remove once per item from B or just the first item from B

To be honest, I agree with your approach:

```red
remove.once  [1 2 [1 2] 3 4 1 2 [1 2] 3 4]  [1 2]; => [[1 2] 3 4 1 2 [1 2] 3 4]
; notice that it removed both 1 and 2, but just the first occurrence.
```

I think it makes much more sense this way.

> If you want to rename, I can. 😉

I think I've already gotten used to it! lol
Seriously, it's fine. 😄

RickBarretto:
Alright! 🙂

RickBarretto:
I'll update this test output. If it runs, I'll open an issue and PR to fix once, and then add a PR for `instance.once`.
As it's on the same file, we would dealing with conflicts what is a boring thing, so I prefer do one at time.

drkameleon:
> I'll update this test output. If it runs, I'll open an issue and PR to fix once, and then add a PR for `instance.once`. As it's on the same file, we would dealing with conflicts what is a boring thing, so I prefer do one at time.

Totally! Don't really worry about the conflicts, I can fix them easily. Let's just get the main thing working and the rest will be done very fast! 😉

RickBarretto:
Why did CI stop?

RickBarretto:
Alright, let's fix once now! 🚀

drkameleon (approved):
Went through your changes... Awesome job, again!

RickBarretto:
The [reverted commit](https://github.com/arturo-lang/arturo/pull/874/commits/6fd02e32171d16b23e862928e820a43853941ba2) was running okay and fixed the `.once`, but I'll do it in another PR.
Also, I know that I can "revert" the commit, but I think it is not the better approach to resolve it.
Now, I'll add some unit tests to cover this new feature.

RickBarretto:
Removing WIP, and read to merge! 🚀
And, also, `.once` behavior was fixed on PR #877

drkameleon (approved):
Looking good! Very nice! 😄 

I'll let the CI builds run after a couple of tiny edits I made, and once they are ready, I'm merging it too!

Good job! 🚀

RickBarretto:
🙂

PR #875, opened 2022-12-15
[Collections/flatten] Make it work with Literal values

### LINKED ISSUE #868: [Collections/flatten] Does not work with Literal values
[Collections/flatten] Does not work with Literal values

This has to be tested out!

https://github.com/arturo-lang/arturo/blob/1136d74ff6685ec5b16c64021b3cf1a0dfc45fa8/src/library/Collections.nim#L482

```text
                    else: push(cleanX[0])

    # TODO(Collections/flatten) Does not work with Literal values
    #  This has to be tested out!
    #  labels: library, bug
    builtin "flatten",
        alias       = unaliased,
        rule        = PrefixPrecedence,
```

17fae356deb9353fcd454bb88450f016fa71855d

RickBarretto:
# Description

Adds literal type support to flatten function.

Fixes #868

On my local build:

image

## Type of change

- [x] Add unit tests

## Requires
- [x] Build update
- [x] Unit test output update


drkameleon:
In this particular case, it seems like the culprit was a missing trailing space, nothing to do with newlines.

And as weird as it sounds, since trailing spaces *do* play an important role in some of the tests, we cannot do it in a different way, unfortunately...

That being said, the PR looks good!

Once the CI builds succeed, I'll merge it too. 🚀

drkameleon:
Ready to merge! 🚀

RickBarretto:
> And as weird as it sounds, since trailing spaces do play an important role in some of the tests, we cannot do it in a different way, unfortunately...

Alright, I will disable the VsCode's trailing spaces for tests folder. :100: 
It explains a lot

PR #879, opened 2022-12-17
Add support of module function for windows

### LINKED ISSUE #878: [Paths/module] module isn't working properly for Windows OS
Initially we are think that were an error from [Collections/split], as related on #872

More detailed what is happening here. (The problem isn't on this library itself), look:

I created a file called `paths.art` with this code:
image

Look what happens when I run it:

image

So, look at lines [1750](https://github.com/arturo-lang/arturo/blob/1136d74ff6685ec5b16c64021b3cf1a0dfc45fa8/src/library/Collections.nim#L1750) and [1797](https://github.com/arturo-lang/arturo/blob/1136d74ff6685ec5b16c64021b3cf1a0dfc45fa8/src/library/Collections.nim#L1797). 
Arturo is using the function `item.split(DirSep))`, `split` is a nim's standard function from `strutils.nim` module, none problem here, and `DirSep` that also is a nim's standard function comes from `osseps.nim` module, none problem here too.

But, let's see what `DirSep` returns for windows systems: 
```
  DirSep* =
    when defined(macos): ':'
    elif doslikeFileSystem or defined(vxworks): '\\'
    elif defined(RISCOS): '.'
    else: '/'
    ## The character used by the operating system to separate pathname
    ## components, for example: `'/'` for POSIX, `':'` for the classic
    ## Macintosh, and `'\\'` on Windows.
```
Yes, it'll split using a `\`. As Arturo's `module` function is returning with bot separators, it won't work. It'll separate just until user's path, and then a big string.

It's a issue related with [`module`](https://github.com/arturo-lang/arturo/blob/master/src/library/Paths.nim#L246) function.

_Originally posted by @RickBarretto in https://github.com/arturo-lang/arturo/issues/872#issuecomment-1353092615_

RickBarretto:
It's not so important because `module` should be replaced on future. But, for while, why not fix it temporarily? 🙂 

Fixes #878

My local build:
image

## Type of change

### DIFF
--- a/src/library/Paths.nim
+++ b/src/library/Paths.nim
@@ -258,7 +258,10 @@ proc defineSymbols*() =
             do module 'html    ; (imports given module)
             """:
                 #=======================================================
-                push(newString(HomeDir & ".arturo/lib/" & x.s & ".art"))
+                when defined(windows):
+                    push(newString(HomeDir & ".arturo\\lib\\" & x.s & ".art"))
+                else:
+                    push(newString(HomeDir & ".arturo/lib/" & x.s & ".art"))
         
         builtin "normalize",
             alias       = dotslash, 


RickBarretto:
🚀

drkameleon (approved):
Since it's looking good, and as you said it's no harm (whether it'll be replaced or not), I'm merging it!

Thanks! 😄

PR #888, opened 2022-12-19
[Collections/split] `.every` is not working properly with String values

### LINKED ISSUE #887: [Collections/split] `.every` is not working properly with String values
**Describe the bug**

> Alright, I discovered another issue. This time not related with literals, but with strings.
> When I run, this:
> ```
>    debug split.every: 3 "Manchester"
>    debug split.every: 3 "Manchesterr"
> ```
> I get this:
> image
>
> It's filling this block with some strange character.
> (Also, I redirected it from bash, so it's not a redirection/conversion error)

The bug was described [in this commentary](https://github.com/arturo-lang/arturo/issues/885#issuecomment-1358069385).
I discovered it when doing some unit-tests.

**To Reproduce**
```
; I will do using string kind literals, but the same is valid for non-literals too
city: "Manchester"
split.every: 3 'city
size city\3
; => 3
```

**Expected behavior**
Size of `city\3` must to be `1`.

**Screenshots**
image
image

**Desktop (please complete the following information):**
 - OS: amd64/windows
 - Version v/0.9.80 b/3311

**Additional context**
[in my commentary](https://github.com/arturo-lang/arturo/issues/885#issuecomment-1358069385) at #885

RickBarretto:
# Description
The current implementation on master is returning strings with same size of `every` attribute.
It's not wrong for strings at the start or middle of the block, and some cases also not wrong at the end.
BUT, for some strings divided by certain numbers, the last item will return with wrong size
So, I fixed it to return the right size.

**Example**
*Current implementation:*
```
; This code
split.every: 3 "Manchester"
; returns a block with:
; [Man che ste r]
; string of size 3, string of size 3, string of size 3 and string of size 3???
```

The real case of this problem is described on issue #887 

*The new implementation*
```
; This code
split.every: 3 "Manchester"
; returns a block with:
; ; [Man che ste r]
; stringsof size 3, string of size 3, string of size 3 and string of size 1
```

The current implementation was putting wrong chars at the last block, since it was pushing a whole `every` size per block without any flexibility.

Fixes #887

## At my local build:
image
image

## Type of change

- [x] Add/Update unit-tests

## Requires
- [ ] Build update
- [x] Unit test output update

### DIFF
--- a/src/library/Collections.nim
+++ b/src/library/Collections.nim
@@ -1785,11 +1785,16 @@ proc defineSymbols*() =
                         var length = InPlaced.s.len
                         var i = 0
 
-                        while i < length:
-                            ret.add(InPlaced.s[i..i+aEvery.i-1])
-                            i += aEvery.i
+                        while i <= length:
+                            if i + aEvery.i <= length:
+                                ret.add(InPlaced.s[i..i+aEvery.i-1])
+                                i += aEvery.i
+                            else:
+                                ret.add(InPlaced.s[i..^1])
+                                i += aEvery.i
 
                         SetInPlace(newStringBlock(ret))
+
                     else:
                         SetInPlace(newStringBlock(toSeq(runes(x.s)).map((x) =>
                                 $(x))))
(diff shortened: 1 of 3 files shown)


drkameleon:
Awesomeness! :)

I just fixed one merge conflict and we are good to go. 😉 

Ready to merge! 🚀

PR #924, opened 2022-12-21
[Collections/rotate] fix in-place behavior for String values

drkameleon:
# Description

The issue exists when the passed String value is in the form of Literal. (in-place `rotate`)

Fixes #920 

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

Issue #920, opened 2022-12-21
[Collections/rotate] is not working for string-literals


RickBarretto:
**Describe the bug**
It's working well for blocks, block-literals and strings itself, but not for string-literals.

**To Reproduce**
```
;; Working well
rotate [2 3 4 5 1] 1
; => [1 2 3 4 5]

;; Working well
a: [ 2 3 4 5 1]
rotate 'a 1
a
; => [1 2 3 4 5]

;; Working well
rotate "rturoA" 1
; => Arturo

;; Broken
b: "rturoA"
rotate 'b 1
b
; => b
```

**Expected behavior**
```
b: "rturoA"
rotate 'b 1
b
; => Arturo
```

**Screenshots**
![image](https://user-images.githubusercontent.com/78623871/208889663-7f544064-1396-495b-8f33-c8465e769383.png)

**Desktop (please complete the following information):**
- os: `amd64/windows`
- version: `v/0.9.80 b/3432`

**Additional context**
- Discovered on PR #910

Issue #912, opened 2022-12-21
[Collections/decouple] is not working with literals


RickBarretto:
**Describe the bug**
decouple is not working with literals. Also, it's having some incongruences, as you can see on the screenshots.

**To Reproduce**

> Note: with `;` are returns from REPL, and `;;` are my commentaries 

```arturo
a: [[1 "one"] [2 "two"] [3 "three"]]

;; With non-literal
decouple a ; => [[1 2 3] [one two three]]
a ; => [[1 one] [2 two] [3 three]]

;; With literal
decouple 'a ; => [[1 2 3] [one two three]]
;; A question about it is: It's really returning something?
a ; => [[1 one] [2 two] [3 three]]
```

**Expected behavior**
```arturo
a: [[1 "one"] [2 "two"] [3 "three"]]

;; With non-literal
decouple a ; => [[1 2 3] [one two three]]
a ; => [[1 one] [2 two] [3 three]]

;; With literal, the same behavior of non-literal
decouple 'a  ;; Does not show anything
a ; => [[1 2 3] [one two three]]
```

**Screenshots**
![image](https://user-images.githubusercontent.com/78623871/208788015-661e7604-6492-40e9-ae42-897ab0d65e4b.png)

Also, another strange beavior:
![image](https://user-images.githubusercontent.com/78623871/208788370-a3d7e4b6-01b4-4ca8-aabd-e5bbb9f41bd2.png)

**Desktop (please complete the following information):**
- os: `amd64/windows`
- version: `v/0.9.80 b/3432`

**Additional context**
- Note: It's working well for non-literal block
- Discovered on PR #910

drkameleon:
> ;; A question about it is: It's really returning something?

Well, it most certainly shouldn't return anything (since it's in-place, after all).

Let me see...

PR #944, opened 2022-12-28
Fix hashing for Literal values

### LINKED ISSUE #142: [Value\hash] Verify hashing is done right
[Value\hash] Verify hashing is done right

https://github.com/arturo-lang/arturo/blob/a971add892fe3d675b3320f356cf2d96179e2a22/src/vm/value.nim#L1400

```text
            result = ""

# TODO(Value\hash) Verify hashing is done right
#  labels: vm,unit-test
proc hash*(v: Value): Hash {.inline.}=
    case v.kind:
        of Null         : result = 0
```

372d19298adc4e945a3632bcaffe16bc123a6cf6

drkameleon:
# Description

There has been a real - albeit only apparent in the REPL - bug that has been affecting the evaluator.

Apparently if we open up the REPL and do something as simple as:

```red
$> b: "str"

$> upper b
=> STR

$> upper 'b
=> STR

$> b
=> str
```

What appears to be happening here is the penultimate command *does* return something (despite it being an in-place argument) and when trying to see the content of our variable (the last command), it seems not to have been changed at all.

**The issue seems simple but in fact it was far more complicated:**

In fact, when the REPL converts the text input (our code) into an executable block, it first has to parse it and evaluate it. I said "evaluate" it. Well, quite. In fact, considering the block a non-dynamic one (which we haven't defined anywhere, but oh well...), it considers that if it has evaluated it before (= to be read as: *converted into bytecode*), the evaluation result would be the same. And, thus, instead of going through the whole bytecode conversion process, it just looks up the given block in a hash of already-evaluated blocks. So, in the end, it's all about hashing the values correctly. 🤔 

**So, what happened here?**

Literals and words were hashed only based on their string value, so `[upper s]` and `[upper 's]` would have the same hash result. In a few words, when the evaluator saw that `[upper 's]`, it thought it had already been evaluated (which was not true ofc) and thus returned the bytecode that was for `[upper s]`... 😢 

**How to solve it**

Hash values properly, in a way that we differentiate between different Value types, despite the fact that their, e.g. string content, *may* end up being identical. 😉 

**Related:**

- #920 
- #924 
- https://github.com/arturo-lang/arturo/issues/912#issuecomment-1361172739

Fixes #142 

## Type of change

- [x] Code cleanup


drkameleon:
Issue to be considered fixed; thanks to @RickBarretto who initially spotted it! 😉 🎉 

-----

Ready to merge! 🚀

PR #945, opened 2022-12-28
Use `OrderedSet`s for methods in Sets module

drkameleon:
# Description

Given the new hashing system, many of the results from the Sets module methods look a bit messed up. At least, we could give the option whether the user doesn't care about the order... but take it for granted that he'd expect the same order and not some random one.

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #968, opened 2023-01-11
[VM/eval] Fix pending issues, add unit-tests & cleanup

drkameleon:
# Description

This PR ended up being far more important than initially planned but it was for a very good reason: there were deep-hidden issues that could *not* have been solved but via a very radical approach.

**So, what does this PR do?**

- Introduce a new intermediate representation (AST), between the parser and the evaluator
- Given this new representation we can:
   - Perform optimizations on the node level (e.g. constant folding, complicated arithmetic simplifications, etc)
   - Optimize the constructs we already were, but in a much safer - and hopefully performant - way
   - All syntactic sugar is/should-be working flawlessly: `->`, `=>`, `|` (and any combinations thereof) ❗ 
- Complete rewrite of the Evaluator (VM/eval), from scratch
- Added more bytecode optimizations *in* the evaluator (e.g. `opStorl`), which was only limited to the now-obsolete optimizer and not even in use before
- The code is far (far...) cleaner and easier to debug and reason about
- Re-organized our opCode set
- Make Built-in functions that correspond to a specific opCode "carry" this opCode as an internal field

Basically, pure bliss! lol 🎉  (seriously now: you cannot imagine how satisfied I am after all of this work!)

Apart from the fixes mentioned below and some general cleanup, this PR - inspired by the awesome work by @RickBarretto - will also attempt to add thorough tests that verify the very important job of the VM's evaluator + AST, as well as revisit previously-written Rosetta Code examples, rewrite any that need rewriting & add the ones that are valid - and perform well enough with the latest version - to our CI builds (= to be executed as part of the Tester tool). 🚀 

-------

- [x] Fixes #799 
- [x] Fixes #802 
- [x] Fixes #803 
- [x] Fixes #804 
- [x] Fixes #817 

-------

## Type of change

- [x] Code cleanup
- [x] Bug fix (non-breaking change which fixes an issue)

PR #969, opened 2023-01-13
Implement AST as intermediate representation

drkameleon:
# Description

**Right now, the process is:**

- the parser ([vm/parse](https://github.com/arturo-lang/arturo/blob/master/src/vm/parse.nim)), well,... parses the given code (= text) into a series of Arturo values (blocks, integer, you name it)
- the resulting block of values is passed to the evaluator ([vm/eval](https://github.com/arturo-lang/arturo/blob/master/src/vm/eval.nim)) which converts it to bytecode
- the bytecode gets executed by our Loop ([vm/exec](https://github.com/arturo-lang/arturo/blob/master/src/vm/exec.nim)) and that's it

**The problem arises in the evaluator:**

At the beginning, things were much simpler, but the more Arturo grows the more this particular part has become quite a mess with various hard-to-debug issues, leaving little room for more optimizations.

**What the idea here is:**

Introduce a new module in the process above; plainly speaking, an intermediate transformation *after* the parsing stage and *before* the evaluation stage. 

And that would be an AST (= [abstract syntax tree](https://en.wikipedia.org/wiki/Abstract_syntax_tree)).

------

Given the complicated (and potentially not even beneficial) nature of this undertaking, this will be marked as highly expirimental (to be read as: it may totally be abandoned and we end up continuing from exactly where we left the project...)

Either way, let's go for it! 🚀 

## Type of change

- [x] Code cleanup


drkameleon:
To be also marked as complete. The final touches will be done in #968.

Either way, things are looking very, very good! 😄 

----

Ready to merge! 🚀

PR #994, opened 2023-02-18
Fix `median` function

RickBarretto:
# Description
> ```red
> print median [1,4,3,2]   ; 3.5 ???????
> ```
> Surely this should be 2.5?
reported [here](https://discord.com/channels/765519132186640445/829324913097048065/1076605657177931817).

`median` isn't sorting the array before take the median itself.

Function was just taking the middle term, even unsorted.

## Type of change


RickBarretto:
> I have heard there are some linear-time algorithms for computing the median that don't require sorting. Might be interesting to look into
> - [Krenium](https://discord.com/channels/765519132186640445/829324913097048065/1076624995217854494)

I'm searching for a new implementation, avoiding the `O( n log n)` performance. I found [this article (Median-finding Algorithm)](https://brilliant.org/wiki/median-finding-algorithm/) at Brilliant.org.

RickBarretto:
I'll probably be afk tomorrow.

BTW, the last commit is working, it just doesn't have the best performance of the world.

I'll try to implement a better soon.

src/library/Statistics.nim
```diff
@@ -134,24 +134,25 @@ proc defineSymbols*() =
...
+
             print median [1 5 2 3 4 7 9 8]
             ; 3.5
         """:
             #=======================================================
-            if x.a.len==0: 
+            if x.a.len==0:
                 push(VNULL)
             else:
+                x.a = x.a.sorted(order = SortOrder.Ascending)
```
drkameleon, line 145:
```suggestion
                let sorted = x.a.sorted(order = SortOrder.Ascending)
```

...and then use `sorted` instead of `x.a`, would be a good idea.

Basically, there is a potential problem with directly sorting and setting `x.a` (haven't tried, you may; but I have a feeling it won't work properly). In a few words, I think the end result of this would be:

```
a: [3 1 2 4]
print median a ; it will print 2.5
print a ; now the array has been sorted, which was not intended after all
```

Not 100% sure, but I think it's quite probable. If I'm wrong, obviously you can ignore this comment 😉

drkameleon:
Awesome and fast job as usual! 🚀 

Please, have a look at my comment to verify we are not actually messing up more things than we need to and we can proceed.

Regarding speed, if there's something that we can do to avoid sorting the list before finding the median, obviously that would imply some performance gain. If not, I guess it's still OK... 😉

src/library/Statistics.nim
drkameleon: Just tested it, and unfortunately I was right. The `median` function now works properly (which is great), but the issue with the passed array being changed is there too:

Screenshot 2023-02-20 at 11 09 40

RickBarretto: I'll solve it soon. At night I'll be home.
You're 100% alright here.
*Mea culpa!*

I missed this part, array should never be modified by this function.

drkameleon: No worries! It's an easy trap, in which I've fallen more times than I can count! haha

It's just that those `x`, `y`, `z` that are auto-magically appear available in built-in function implementation are nothing but pointers. So, we may use them, but not change them. (or anything they point to...) 😉

src/library/Statistics.nim
```diff
@@ -134,24 +134,25 @@ proc defineSymbols*() =
...
             print median [1 5 2 3 4 7 9 8]
             ; 3.5
         """:
             #=======================================================
-            if x.a.len==0: 
+            if x.a.len==0:
                 push(VNULL)
             else:
+                let sorted = x.a.sorted(order = SortOrder.Ascending)
                 let first = x.a[(x.a.len-1) div 2]
```
drkameleon, line 146:
```suggestion
                let first = sorted[(sorted.len-1) div 2]
```

src/library/Statistics.nim
```diff
@@ -134,24 +134,25 @@ proc defineSymbols*() =
...
                 let second = x.a[((x.a.len-1) div 2)+1]
```
drkameleon, line 147:
```suggestion
                let second = sorted[((sorted.len-1) div 2)+1]
```

src/library/Statistics.nim
```diff
@@ -134,24 +134,25 @@ proc defineSymbols*() =
...
 
                 if x.a.len mod 2 == 1:
```
drkameleon, line 149:
```suggestion
                if sorted.len mod 2 == 1:
```

Since we now have `sorted`, better use this one. (OK, it's a bit of an overkill to use for `.len` as well, but it looks cleaner I guess). In the case, of `first` and `second`, it's a must, since x.a is not sorted 😉

RickBarretto: Mea culpa, I forgot to update the variables, since I committed the code from Github's fast commit.

RickBarretto:
Hum, something is wrong here for FreeBSD. Lemme see...

RickBarretto:
I committed a broken code, I know.
It's compiling, but it's not working for arrays greater than `tiny`, it doesn't show any special message, just a Windows' pop-up saying "arturo.exe stopped to work.".

I think I might need some help here, @drkameleon.

RickBarretto:
I'll see what is happening with [Console Log](https://nim-lang.org/docs/logging.html#ConsoleLogger)

RickBarretto:
I think I found a infinity recursion... LOL!
image

RickBarretto:
No more infinity recursion... (the problem was on `distribute`, so I made my own)
BUT, I must test it before, check if it's returning me the right value for every possible case...

drkameleon:
This is looking very good, honestly. Far better than I expected for something as trivial as a `median` function.

Great job! 😉 

Let me know when you consider it ready and I'll have another look into it so that we can merge it! 🚀

RickBarretto:
It'll fail...
The most of the tests passed, but, this `[9, 6, 12, 5, 17, 3, 9, 5, 10, 2, 8, 7]` in special is returning `null`... 
Why? IDK... I'm trying to figure out.

```
$> median [9, 6, 12, 5, 17, 3, 9, 5, 10, 2, 8, 7]
SIGSEGV: Illegal storage access. (Attempt to read from nil?)
```

RickBarretto:
[The current code on Playground](https://play.nim-lang.org/#ix=4oTq)

It returns me:
```nim
/usercode/in.nim(73)     in
/usercode/in.nim(65)     medianOfMedians
/usercode/in.nim(67)     medianOfMedians
/usercode/in.nim(44)     medianOfMedians
/playground/nim/lib/system/fatal.nim(54) sysFatal
Error: unhandled exception: index 3 not in 0 .. 2 [IndexDefect]
```

drkameleon:
> [The current code on Playground](https://play.nim-lang.org/#ix=4oTq)
> 
> It returns me:
> 
> ```nim
> /usercode/in.nim(73)     in
> /usercode/in.nim(65)     medianOfMedians
> /usercode/in.nim(67)     medianOfMedians
> /usercode/in.nim(44)     medianOfMedians
> /playground/nim/lib/system/fatal.nim(54) sysFatal
> Error: unhandled exception: index 3 not in 0 .. 2 [IndexDefect]
> ```
> 
> This function should return `8` as value.

Let me have a look...

Inside info: This `indexDefect` type of compiler exception can be quite nightmarish, and most likely it's nothing that big of a mistake on your side... Since, it's compiler related, I would look into things like `import`s, etc, but still not really remember the last time I had such an issue (and how I solved it).

One similar case of me struggling with the exact same kind of error: https://forum.nim-lang.org/t/8976 ... 🤔

drkameleon:
> @drkameleon I think I found the error, and I'm fixing it...

Let's see... I was looking at it *right now*, but I'm quite baffled tbh...

RickBarretto:
Humm... I'll try again...

RickBarretto:
Alright, I don't know what is broken here.
I'll try with the original from brilliant and see if it's not the algorithm itself.

RickBarretto:
It's the algorithm itself, tried with Python:

image

RickBarretto:
I'm testing a new approach based on this article: [My Favorite Algorithm: Linear Time Median Finding](https://rcoh.me/posts/linear-time-median-finding/).

It'[s working well](https://play.nim-lang.org/#ix=4oUN).

RickBarretto:
@drkameleon 
Aaaannnd, working! 🙂

drkameleon (approved):
Looking great! Very, *very* good job as usual!

Also: good idea to have a separate helper module for Statistics-related methods + the unit-tests (which are always a great idea, although I tend to forget about them...)

drkameleon:
I'm going to ignore the supposedly-failing CI action (it's the usual `failed to fetch <some_deb_package>` nonsense, in Github actions...), so let's just move on! 😉 

Ready to merge! 🚀

PR #1008, opened 2023-02-26
Add unit-tests for Files Library

RickBarretto:
# Description

Adds unittests for Files Library.
Notes were added if founded strange behaviors or bugs.

## Type of change

- [ ] Update/Add unit-tests


RickBarretto:
Notes:

- `delete.directory` deletes directory itself, but what if I want to just remove the its content?
  - `.content` or `.within` attributes would be a good idea.
- I'm using `write` to create some test assets because I can't do bash's redirections on `execute` function. Is it a bug?

RickBarretto:
Notes:

- `permissions.set` raises an error: `SIGSEGV: Illegal storage access. (Attempt to read from nil?)`

RickBarretto:
Notes:

- What is the difference between `read` and `read.lines`?
- What is the difference between `read` and `read.file`?
- How can I test `read.bytecode`?

drkameleon:
>   * `.content` or `.within` attributes would be a good idea.

I would go for `.content` or `.contents`, both sound very self-explanatory. E.g. `delete.directory.content` or `delete.directory.contents` - looks super nice! :)


> * I'm using `write` to create some test assets because I can't do bash's redirections on `execute` function. Is it a bug?

I'm not really sure what you mean. Can you elaborate a bit?

drkameleon:
> Notes:
> 
> * `permissions.set` raises an error: `SIGSEGV: Illegal storage access. (Attempt to read from nil?)`

That's bad. I'll test it a bit (which I haven't, to be perfectly honest) and we could open an issue about it... Good catch! 😉

RickBarretto:
> I'm not really sure what you mean. Can you elaborate a bit?

I can't write files using bash redirections:

```bash
execute "echo 'hello, world' > file.txt"
```

Or even
```bash
execute "echo 'hello, world' | tee file.txt"
```

drkameleon:
> * What is the difference between `read` and `read.lines`?

Let's say we have a file with the following contents:

```
one
two
three
four
```

- `read` takes a filepath and reads the whole file content as a string (*one* string):

Screenshot 2023-02-27 at 11 03 51

- `read.lines` does the same, only it reads the file as an array of lines/strings:

Screenshot 2023-02-27 at 11 05 16

Why? The reason is simple. If you read a file, and want to go through its different lines, you'd first have to do sth like `split.lines read ...`. With `read.lines`, you can directly do it like `loop read.lines "myfile" 'line -> ...` 😉 

-----

> * What is the difference between `read` and `read.file`?

`read` is probably the most... underestimated function in the whole standard library, since it does much more than it appears to. So, what does `read` take as a parameter? A string. But *what* is this string? The truth is... it can be anything. If you pass a URL and Arturo recognizes it as such, it will actually download the text file/page in the background and perform whatever you want with it. If it's a file path and the filepath exists, it'll do the same with the file contents. Now, if none of this is valid, it will treat it as a regular string...

So, let's take for example `read.json`. This configuration reads a JSON structure and returns an Arturo object/dictionary/array/etc. So, you can actually do any of these things:

- `read.json "[1,2,3]"`
- `read.json "myjsonfile.json"`
- `read.json "http://somedomain.com/some.json"`

Now, the problem with this is... that if the passed filepath doesn't exist, or you've made a typo, instead of showing an error... it will read this as a normal string. Example:

- you want to `read.file "myfile.txt"` (so that it returns the file's *contents*)
- you actually write, by mistake or whatever, `read.file "myfil.txt"` (in that case, instead of throwing an error, since it doesn't know you meant a file, and that file doesn't exist, the call will return "myfil.txt" (the string itself) which is not what you intended.

In that case, `read.file` forces a string to be read as a filepath. And if it doesn't exist, it'll throw an error. 😉 

> * How can I test `read.bytecode`?

Basically, it can work with filepath's produced via `arturo -c somescript.art`. This creates a bytecode file. Which you can then "load" or `read` via `read.bytecode` 🚀

drkameleon:
> > I'm not really sure what you mean. Can you elaborate a bit?
> 
> I can't write files using bash redirections:
> 
> ```shell
> execute "echo 'hello, world' > file.txt"
> ```
> 
> Or even
> 
> ```shell
> execute "echo 'hello, world' | tee file.txt"
> ```

Hmmm... I think this might be more complicated. I'm not sure pipes and redirections, Bash-style, would work via `execute` but I'll have to investigate a bit more 😉

drkameleon:
All that being said: *great* job! 😉  

What you're doing with the unit-tests, in general, is very very useful and of extreme relevance (plus, it helps pick issues that may have been ignored)

Keep going! 🚀

RickBarretto:
Thanks a lot for explanation, @drkameleon !

I'll cover `read.bytecode` and `read.file`, and makes sense why execute don't allow redirections/pipes.

> read.lines does the same, only it reads the file as an array of lines/strings:

But it was returning me as a string instead.
I got it why this attribute is here, but the returned value was wrong.

---

Also, is there a way to read just some parts of a file? I mean, if you load all file inside a string or block of strings you are allocating it to RAM. That means that we should avoid load big files, isn't?

drkameleon:
> > read.lines does the same, only it reads the file as an array of lines/strings:
> 
> But it was returning me as a string instead. I got it why this attribute is here, but the returned value was wrong.

I'll double-check this right now, although - given that's it's used quite a lot already, and in different places - there shouldn't be any issue. Let's see... 🤔 

> Also, is there a way to read just some parts of a file? I mean, if you load all file inside a string or block of strings you are allocating it to RAM. That means that we should avoid load big files, isn't?

That's a great idea and something that I have contemplated.

Basically, my idea was even more ambitious:

- have the possibility of reading a file as data, e.g. an `:integer`, etc
- also, have the option to treat a file as a stream

I'm totally in; I just think it would need some very careful planning 😉

RickBarretto:
Alright! 🙂

Btw, I must re-test all my "bugs reports". I'm going to do it at the end of this PR.

Re-test, everything and check if it's every okay.

drkameleon:
> But it was returning me as a string instead. I got it why this attribute is here, but the returned value was wrong.

Unfortunately, I'm not able to reproduce this. If you have some concrete example where it returns a wrong value, I could perhaps manage to locate it... 🤔

RickBarretto:
This code here:
image
Is resulting into an empty zip file:
image

Have I done something wrong here, or is it a bug?

drkameleon:
> Is resulting into an empty zip file:
> ...
> Have I done something wrong here, or is it a bug?

This is what I just tried (simplest scenario of all) - in the main `arturo/` folder:

```red
zip "dest.zip" ["README.md"]
```

- the `dest.zip` file is generated correctly
- the contents of the .zip are correct

Now, your example not working would have me thinking 2 different things:
- either it is a bug, that has to do with the different paths being passed to `zip` (I would experiment a bit, e.g. trying to save the zip in the same folder, etc)
- or, Windows auto-matically filters some of the contents of the zip (I'm not an expert at this, but it did happen to me with some of Arturo's build zips...)

RickBarretto:
> but it did happen to me with some of Arturo's build zips...

Ohh, Windows... My passion...

RickBarretto:
I found what is happening here:

When I run this:
```
$> zip "dest.zip" ["src"]
```

I got a:

image

So, maybe it's not working for folders. I'm going to test further this.

drkameleon:
> I found what is happening here:
> 
> When I run this:
> 
> ```
> $> zip "dest.zip" ["src"]
> ```
> 
> I got a:
> 
> image
> 
> So, maybe it's not working for folders. I'm going to test further this.

Hmm... Interesting... 🤔 

In theory, it shouldn't matter. I'll have to have a look into the implementation of the zip component.

drkameleon:
Let's see...

The whole things starts here: https://github.com/arturo-lang/arturo/blob/master/src/library/Files.nim#L572

And the main implementation for this `zip` call is here: https://github.com/arturo-lang/arturo/blob/master/src/extras/miniz.nim#L853-L861

Now, looking at it, there is one particular line that leads me to believe that something may not be done right: https://github.com/arturo-lang/arturo/blob/master/src/extras/miniz.nim#L858

Obviously, if the function works only for simple files and not for files-in-folders (or folders) - which would be quite a normal scenario, we'll have to create an issue for that as a bug.

RickBarretto:
Missing tests:
- [ ] `symlink`: how can I properly test it?
- [ ] `read.bytecode`: I can't delete the `hello.art.bcode` using `execute` or `delete`, what can break the next tests
- [ ]  `zip` & `unzip` for folders
- [ ] `permissions.set`: returns a SIGSEGV

RickBarretto:
The test will break because the MSYS's bash redirection isn't writing on right format...

RickBarretto:
@drkameleon can you try to generate this `res` file?

drkameleon:
> `symlink`: how can I properly test it?

I think the easiest way would be for me to add a `.symlink` option to the already existing [`exists?`](https://github.com/arturo-lang/arturo/blob/master/src/library/Files.nim#L117-L140) function, using Nim's [`symlinkExists`](https://nim-lang.org/docs/os.html#symlinkExists%2Cstring). Give me some time and you may pull directly from the master branch 😉 

> `read.bytecode`: I can't delete the hello.art.bcode using execute or delete, what can break the next tests

That's weird. So, you create this `hello.art.bcode` and `delete` doesn't work on it? 🤔 

> `zip` & `unzip` for folders

I'll open an issue for this. I think it's officially a... bug (unfortunately). But good job, since we caught it! 🚀 

> `permissions.set`: returns a SIGSEGV

That's another bug I guess. I haven't tested this particular functionality a lot.. so it looks like it's broken. I'm having a look into it!

drkameleon:
> @drkameleon can you try to generate this `res` file?

I'll clone and try to... Just give me a few minutes! 😉

drkameleon:
**Generated .res** (disclaimer: I cannot confirm what is correct and what isn't; at least for now):

https://gist.github.com/drkameleon/e0a2285a464b79b3a9fd55674bc9a79d

RickBarretto:
I'm going to have a breakfast and check! 😉

RickBarretto:
I must replace all `execute "ls ..."` commands. It's printing with different formatations on Msys and MacOS. Look:

image

I'm going to rewrite all of these, and use Arturo's functions. The problem is that it turns these File's functions dependent of each other...

RickBarretto:
Since, `exists?`, `write`, `read` and `delete` are working well, I don't thing it's gonna be a problem.
But I need your opinion about it.

drkameleon:
> I'm going to rewrite all of these, and use Arturo's functions. The problem is that it turns these File's functions dependent of each other...

I'm not sure I get the last part.

Regarding the first part, I think I would personally use Arturo's own [**`list`**](https://arturo-lang.io/documentation/library/paths/list/).

For example, `ls temp/dest/** --recursive` would correspond to:

```red
list.recursive "temp/dest"
```

RickBarretto:
> I'm not sure I get the last part.

I was using `execute` to avoid to use Arturo's built-in functions.
Let's say that `list` is broken... It'll break other tests that uses `list` internally, for instance.

But, I think we can use built-in functions as well.

drkameleon:
> > I'm not sure I get the last part.
> 
> I was using `execute` to avoid to use Arturo's built-in functions. Let's say that `list` is broken... It'll break other tests that uses `list` internally, for instance.
> 
> But, I think we can use built-in functions as well.

I totally see your point.

Personally, I *would* use built-in functions - I think it's inevitable basically + it makes our life easier, with a tiny asterisk... meaning, if that if something actually breaks at some point, to have that in mind (or as a comment). But, other than that, I would go ahead... 😉

RickBarretto:
Apparently, the problem with `zip`/`unzip` is the fact of `zip` can't work with paths.
That is the reason why it works for the example of "README.md", I'm not passing a path here, just the file itself.
While for the others examples it breaks.

For instance,
This:
```red
topic "zip & unzip - arturo.nim"
do [
    dir: createTestFolder "zip-unzip-src"

    ;; outputs me an empty .zip
    zip ~"|dir|/dest.zip" [~"src/arturo.nim"]
    ensure -> exists? ~"|dir|/dest.zip"
    print "zipped as dest.zip"

    ;; Tries to unzip an empty file
    unzip ~"|dir|" ~"|dir|/dest.zip"
    ensure -> exists? ~"|dir|/arturo.nim"

    print "unzipped as arturo.nim"
]
```

Returns me it:
```red
>> zip & unzip - arturo.nim

zipped as dest.zip
>> Assertion | File: lib.files.art
       error | Line: 647
             |
             | [exists? ~ "|dir|/arturo.nim"]
```

The first `ensure` is ok, it's writing the zip file, but the second... Nope. 
And it isn't a `unzip` fault, look:

```red
topic "zip & unzip - README"
do [
    dir: createTestFolder "zip-unzip-readme"

    ;; outputs me an empty .zip
    zip ~"|dir|/dest.zip" [~"README.md"]
    ensure -> exists? ~"|dir|/dest.zip"
    print "zipped as dest.zip"

    ;; Tries to unzip an empty file
    unzip ~"|dir|" ~"|dir|/dest.zip"
    ensure -> exists? ~"|dir|/README.md"

    print "unzipped as README.md"
]
```

Returns me it:

```red
>> zip & unzip - README

zipped as dest.zip
unzipped as README.md
```

---
Attachments:

image
image

drkameleon:
@RickBarretto I have a weird idea as an experiment... if you want, you can try it out.

At https://github.com/arturo-lang/arturo/blob/master/src/extras/miniz.nim#L858

What if we change this:

```
discard pZip.mz_zip_writer_add_file(f.extractFileName.cstring, f.extractFileName.cstring, comment, 0, cast[mz_uint](MZ_DEFAULT_COMPRESSION))
```

to

```
discard pZip.mz_zip_writer_add_file(f.cstring, f.extractFileName.cstring, comment, 0, cast[mz_uint](MZ_DEFAULT_COMPRESSION))
```

I'll properly do in it a different PR I guess, but I think this tiny fix could highlight the underlying issue. Basically, the first parameter to [`mz_zip_writer_add_file`](https://cpp.hotexamples.com/examples/-/-/mz_zip_writer_add_file/cpp-mz_zip_writer_add_file-function-examples.html?utm_content=cmp-true) is the *src* file, while the second is the *dest*ination (in our zip). I think its this first `.extractFileName` that messes the whole thing up...

RickBarretto:
A lot of tries, but I can't test `read.bytecode` it's breaking other tests that isn't broken, returning `SIGSEGV: Illegal storage access. (Attempt to read from nil?)`.
The commented code committed, for instance, breaks `read.file`.
And more, it's returning `:binary` type instead of `:bytecode`.

I could test it right on REPL, but couldn't do it using script.

RickBarretto:
BTW, my MSYS isn't redirecting right. And CI is breaking over and over...
But, I think I finished everything here.

RickBarretto:
And, about CI. It's a quite strange...

- FreeBSD [failed on Setup & Build](https://github.com/arturo-lang/arturo/actions/runs/4297045836/jobs/7489531547)
- Ubuntu [failed on Install dependencies](https://github.com/arturo-lang/arturo/actions/runs/4297045827/jobs/7489531333)

Windows & MacOS is something with enconding or CRLF/LF, IDK...

RickBarretto:
Note: I realized that this branch is not updated with main.

RickBarretto:
> What if we change this:
> 
> ```
> discard pZip.mz_zip_writer_add_file(f.extractFileName.cstring, f.extractFileName.cstring, comment, 0, cast[mz_uint](MZ_DEFAULT_COMPRESSION))
> ```
> 
> to
> 
> ```
> discard pZip.mz_zip_writer_add_file(f.cstring, f.extractFileName.cstring, comment, 0, cast[mz_uint](MZ_DEFAULT_COMPRESSION))
> ```

I tried it locally, but it isn't working anyway... The `zip` file continues empty... 🙁

drkameleon:
> And, about CI. It's a quite strange...
> 
> * FreeBSD [failed on Setup & Build](https://github.com/arturo-lang/arturo/actions/runs/4297045836/jobs/7489531547)
> * Ubuntu [failed on Install dependencies](https://github.com/arturo-lang/arturo/actions/runs/4297045827/jobs/7489531333)

This is a constant pain. It has nothing to do with us. It's sth with the Github Actions (and one of the reasons why I have contemplated alternatives...)

> Windows & MacOS is something with enconding or CRLF/LF, IDK...

Hmm... I'll look into it!

drkameleon:
> I tried it locally, but it isn't working anyway... The `zip` file continues empty... 🙁
 
 I would say you shouldn't bother about this. The bug seems quite serious and - apparently - it cannot be solved just like that.

So... I think it would be better if we solved it properly (there is already an issue opened for that, so that we can track it down) and *then* we could add all the relevant unit-tests here. But for the purposes of this PR, I wouldn't worry too much! 😉

drkameleon:
**UPDATE**

Regarding the different failing CI actions:

- The FreeBSD one seems to be quite random, so I don't know what to say about this (but it doesn't seem to be such a constant issue, so let's pretend to ignore it for now...)
- Regarding the failing Linux builds, since I had fixed something related to different workflows in one of the pending PRs (which I just merged), if you can sync with the master branch one more time,  think we could at least see what it is that is really failing 😉

RickBarretto:
> I would say you shouldn't bother about this. The bug seems quite serious and - apparently - it cannot be solved just like that.

> The FreeBSD one seems to be quite random

Alright, no problem!

---

> read.bytecode could be one of those things that either have a bug, or I forgot to updated their implementation.

Okay, let's left this unit-test commented for while, so

---

>  if you can sync with the master branch one more time, think we could at least see what it is that is really failing 😉

Alright, I can! 😉

drkameleon:
I just did a diff check between the Linux output (which I consider the "safest" of all) and the expected output at .res:

https://www.diffchecker.com/rL9wIKDN/

The main things I notice, at first sight:

- `permissions` see different... permissions
- there are consistent discrepancies when reading any format practically: Toml, Yaml, Markdown, etc (and since in the CI action it returns the content itself, it means it doesn't pick the path correctly 😉 - remember the `.file` option I was talking about, that's exactly what the purpose is)
- I can see a couple of the usual `/` vs `\` in paths...

RickBarretto:
What is breaking here:

image
image
image
image
image

drkameleon:
> > I just did a diff check between the Linux output (which I consider the "safest" of all) and the expected output at .res:
> > https://www.diffchecker.com/rL9wIKDN/
> 
> Wow, I'll use this tool now, I was seeing things manually...

The truth is I used to do it manually too, until your mega-PR (with the collections-related unit-tests), where I kept looking and looking for what was going on and I went crazy... 🤣

drkameleon:
> What is breaking here:
> 
> image image image image image

I'm suspecting this `inspect read.html ~"|dir|/test.html"` (and all related comments) and it has to do with the weird way Arturo handles paths right now... Would `inspect read.html relative ~"|dir|/test.html"` work?

I'd give it a try for sure...

RickBarretto:
> Would inspect read.html relative ~"|dir|/test.html" work?

Yes, I can! But with the original or the modified `inspect`?


I'm using a custom `inspect` command, since I thought that the differences were the fact that the original `inspect` returns a colored output.
So, my custom inspect is just this:

```red
inspect: $[T][
    print T
    print type T
]
```

RickBarretto:
Also, I updated my local branch and re-built arturo.
For some reason, the same code returns me a `:dictionary` for `read.html`.

I'll try your approach and see what the CI returns.

RickBarretto:
`inspect read.html relative ~"|dir|/test.html"` returns: `tests\unittests\temp\read-html\test.html :string`

RickBarretto:
Running with `inspect read.html.file relative ~"|dir|/test.html"` I got a:

```
>> Runtime | File: lib.files.art
     error | Line: 416
           | 
           | file not found: tests/unittests/temp/read-html/test.html
```

drkameleon:
> Also, I updated my local branch and re-built arturo. For some reason, the same code returns me a `:dictionary` for `read.html`.
> 
> I'll try your approach and see what the CI returns.

`read.html` should return a dictionary indeed (the "parsed" HTML tree), so this sounds quite correct

drkameleon:
> Running with `inspect read.html.file relative ~"|dir|/test.html"` I got a:
> 
> ```
> >> Runtime | File: lib.files.art
>      error | Line: 416
>            | 
>            | file not found: tests/unittests/temp/read-html/test.html
> ```
> 
> What is quite strange since it pass on `ensure -> exists? ~"|dir|/test.html"` at line 412.

Obviously, this shows why `read` or `read relative` doesn't really work as it should, path-wise. Noted! (And thanks for reminding me that this *has* to be resolved once and for all 😉 )

Now... if `read` returns the string... it means that it doesn't consider it a file. So, bingo!

Let me clone myself and I'll try to figure out what is going on... 🤔

drkameleon:
So, I just cloned your branch and try running the whole thing (on macOS)...

I was at the root directory (@ arturo/) and running the script like `arturo tests/unittests/lib.files.art` - I'm saying that since it *must* make some difference.

Initially, I got an error like:

```
>> Runtime | File: lib.files.art
     error | Line: 416
           | 
           | file not found: tests/unittests/temp/read-html/test.html
```

Then I just removed this `relative` (that *I* told you to include there) and... bingo!

Apparently, everything works fine. Even Toml, Yaml and all of that.

I'll now try to save the output as `.res` and see if when run via `tools/tester.art` (our tester script), the result is identical. (Obviously, it *should*, unless there is a bug with the paths - which, I think, there is...)

drkameleon:
So... update, and even weirder... 

I copy-pasted and saved the output from `arturo tests/unittests/lib.files.art` to the corresponding `tests/unittests/lib.files.res` and the tester tools works absolutely fine. 😲 

So... I'm thinking...

If anything breaks in the CI action builds, it must be something to do with OS differences... 

For reference, this is the `.res` it produces when I run it... (just with that one `relative` removed):

https://gist.github.com/drkameleon/5b5b854294d8da885e7a477165b9a4a7

tests/unittests/lib.files.art
```diff
@@ -0,0 +1,934 @@
...
+    dir: createTestFolder "read-json"
+
+    write ~"|dir|/test.json" {
+        {
+            "name": "Arturo",
+            "version": "0.9.83",
+            "build": "b/12",
+            "platform": "amd/win10"
+        }
+    }
```
drkameleon, line 338:
```suggestion
    jsonSample: {
        {
            "name": "Arturo",
            "version": "0.9.83",
            "build": "b/12",
            "platform": "amd/win10"
        }
    }
    write ~"|dir|/test.json" jsonSample
    
    inspect read.json jsonSample
```

Also - not that it's *that* important, but since you've done it all in such a nice and organized way - since `read` (or `read.json`, `read.html`, etc for that matter) also work directly on string values, you could also directly test the string value and... then when it's reading from a file, as you already do. 😉 

Just an additional idea - but I'd say we'd better manage to get it to pass the tests first, as it is.

RickBarretto: Alright! 😉

RickBarretto:
> https://gist.github.com/drkameleon/5b5b854294d8da885e7a477165b9a4a7

Alright, there are some differences between yours and mine. [look](https://www.diffchecker.com/GcAIocCw/)
Some difference are the behavior itself, others are spaces at the final of line. But, I copied directly from terminal (without redirection), so mine doesn't have white spaces.

---

Yours at line 40-58:
```
>> permissions

[ :dictionary
        user    :        [ :dictionary
                read     :                true :logical
                write    :                true :logical
                execute  :                false :logical
        ]
        group   :        [ :dictionary
                read     :                true :logical
                write    :                false :logical
                execute  :                false :logical
        ]
        others  :        [ :dictionary
                read     :                true :logical
                write    :                false :logical
                execute  :                false :logical
        ]
]
```

Mine:
```
>> permissions

[ :dictionary
        user    :        [ :dictionary
                read     :                true :logical
                write    :                true :logical
                execute  :                true :logical
        ]
        group   :        [ :dictionary
                read     :                true :logical
                write    :                true :logical
                execute  :                true :logical
        ]
        others  :        [ :dictionary
                read     :                true :logical
                write    :                true :logical
                execute  :                true :logical
        ]
]
```

---

Yours at line 95-118:

```
>> read.csv

[ :block
        [ :block
                language :string
                 version :string
                 platform :string
        ]
        [ :block
                Arturo :string
                 0.9.83 :string
                 win10 :string
        ]
        [ :block
                Python :string
                 3.9 :string
                 gnu/linux :string
        ]
        [ :block
                Ruby :string
                 3.2.1 :string
                 macOS :string
        ]
]
```

Mine:

```
>> read.csv

[ :block
        [ :block
                language; version; platform :string
        ]
        [ :block
                Arturo; 0.9.83; win10 :string
        ]
        [ :block
                Python; 3.9; gnu/linux :string
        ]
        [ :block
                Ruby; 3.2.1; macOS :string
        ]
]
```

---
Yours at lines 387-394:

```
>> hidden?

false
false
true
true
true
true
```

Mine:
```
>> hidden?

false
false
false
false
false
true
```

tests/unittests/lib.files.art
RickBarretto: Replaced! at [97f911c](https://github.com/arturo-lang/arturo/pull/1008/commits/97f911c709d7714ddad95ff725af1803f69cde68)

RickBarretto:
I'll try just a thing. Suppress these 3 tests for while and try to generate a `res` file copying directly from terminal.

drkameleon:
> > https://gist.github.com/drkameleon/5b5b854294d8da885e7a477165b9a4a7
> 
> Alright, there are some differences between yours and mine. [look](https://www.diffchecker.com/GcAIocCw/) Some difference are the behavior itself, others are spaces at the final of line. But, I copied directly from terminal (without redirection), so mine doesn't have white spaces.
> 
> Yours at line 40-58:
> 
> ```
> >> permissions
> 
> [ :dictionary
>         user    :        [ :dictionary
>                 read     :                true :logical
>                 write    :                true :logical
>                 execute  :                false :logical
>         ]
>         group   :        [ :dictionary
>                 read     :                true :logical
>                 write    :                false :logical
>                 execute  :                false :logical
>         ]
>         others  :        [ :dictionary
>                 read     :                true :logical
>                 write    :                false :logical
>                 execute  :                false :logical
>         ]
> ]
> ```
> 
> Mine:
> 
> ```
> >> permissions
> 
> [ :dictionary
>         user    :        [ :dictionary
>                 read     :                true :logical
>                 write    :                true :logical
>                 execute  :                true :logical
>         ]
>         group   :        [ :dictionary
>                 read     :                true :logical
>                 write    :                true :logical
>                 execute  :                true :logical
>         ]
>         others  :        [ :dictionary
>                 read     :                true :logical
>                 write    :                true :logical
>                 execute  :                true :logical
>         ]
> ]
> ```
> 
> Yours at line 95-118:
> 
> ```
> >> read.csv
> 
> [ :block
>         [ :block
>                 language :string
>                  version :string
>                  platform :string
>         ]
>         [ :block
>                 Arturo :string
>                  0.9.83 :string
>                  win10 :string
>         ]
>         [ :block
>                 Python :string
>                  3.9 :string
>                  gnu/linux :string
>         ]
>         [ :block
>                 Ruby :string
>                  3.2.1 :string
>                  macOS :string
>         ]
> ]
> ```
> 
> Mine:
> 
> ```
> >> read.csv
> 
> [ :block
>         [ :block
>                 language; version; platform :string
>         ]
>         [ :block
>                 Arturo; 0.9.83; win10 :string
>         ]
>         [ :block
>                 Python; 3.9; gnu/linux :string
>         ]
>         [ :block
>                 Ruby; 3.2.1; macOS :string
>         ]
> ]
> ```
> 
> Yours at lines 387-394:
> 
> ```
> >> hidden?
> 
> false
> false
> true
> true
> true
> true
> ```
> 
> Mine:
> 
> ```
> >> hidden?
> 
> false
> false
> false
> false
> false
> true
> ```

The permissions thing + what is hidden and what isn't looks too OS-dependent.

Regarding the CSV, ... oops... I forgot to mention a tiny edit I made (instinctively, but still).

Since, CSV stands for Comma-Separated Values, the "original" idea is to recognize, well... comma-separated values. So, I just replaced the semicolons (`;`) in your example input with commas (`,`), so the CSV is instantly parsed better.

Having said that, I guess we should add support for custom CSV delimiters (like semicolons in your example), which despite the naming thing I mentioned could be very common as well. Noted! :)

RickBarretto:
> Since, CSV stands for Comma-Separated Values, the "original" idea is to recognize, well... comma-separated values. So, I just replaced the semicolons (;) in your example input with commas (,), so the CSV is instantly parsed better.

Ohhhh, got it!

RickBarretto:
> Having said that, I guess we should add support for custom CSV delimiters (like semicolons in your example), which despite the naming thing I mentioned could be very common as well. Noted! :)

Good idea, I always use `;` semi-colons to it instead of commas `,`, so, my bad...

drkameleon:
Not really an issue! Obviously, we should be able to support everything. And semicolons *is* indeed very common.

So, to-do! :)

-----

P.S. I'm going offline, since it's getting close to 7PM here and I'm totally exhausted. But I'll be watching 😉

RickBarretto:
> P.S. I'm going offline, since it's getting close to 7PM here and I'm totally exhausted. But I'll be watching 😉

Sure! Have a good rest!

drkameleon:
The challenge I see from the above list is to make `read.toml`, etc just work.

I think it should be doable. Let's see... 🤔 

The rest has already been considered a bug or an existing issue, and will be dealt with.

All in all, great work! I think that - despite having chosen one of the trickiest... modules to write unit-tests for (I could definitely pick much easier ones, trust me! lol) - you have done a very neat job here! 🚀 😉 

Now, let me see if I can get those `read.sth` work on CI...

drkameleon:
I've just noticed something very interesting:

- `read.csv` with a filepath seems to be working fine
- but the rest doesn't

So, it doesn't have to do with filepaths in general. It's something specific about how e.g. `read.html` handles the path...

Very interesting!

drkameleon:
Geeeeeez! I spotted it. 🎉 

Basically, the reasons why `read.html` etc breaks in quite simple:

- one of the things that are disable in MINI builds is all these parsers (html, xml, toml, etc)
- disabled means the `.html`, `.xml`, etc options will be ignored
- thus the passed string is read as... nothing and returned as-is

So, how to solve it?

As with all things that cannot be run in the MINI builds, we'll just "skip" the test for MINI builds by adding `lib.files` to the exception list here:

https://github.com/arturo-lang/arturo/blob/master/tools/tester.art#L19-L50

Just add `"lib.files"` along with `"branching"` etc... and if you want, try re-enabling the tests that were not working (basically the parser-related ones). And... I *think* it will all work fine. (Since the MINI builds failing meant that the whole workflow here @ github was bound to be terminated) 🚀

RickBarretto:
Huummm, got it! I'm going to enable these tests again. And disable it for MINI builds.

RickBarretto:
Why is it running in Mini yet?

https://github.com/arturo-lang/arturo/actions/runs/4316514399/jobs/7532384578
image

drkameleon:
Let's see if this tiny fix works... 😉

RickBarretto:
> Let's see if this tiny fix works... 😉

We had two problems here.
My fault in both... LOL

1. I added the extension [here](https://github.com/arturo-lang/arturo/pull/1008/commits/2ffb3c1d475eff9d7b24ce6ed00ecd52300827a7): `"lib.files.art"`
2. I wrote a tiny bug [here](https://github.com/arturo-lang/arturo/pull/1008/commits/997a1cbbb95edf55412dc3bace91f12af6496907): `write ~"|dir|/test.json" write ~"|dir|/test.json" jsonSample`

OMG, I think I would run well now. 🙂

drkameleon:
> > Let's see if this tiny fix works... 😉
> 
> We had two problems here. My fault in both... LOL
> 
> 1. I added the extension [here](https://github.com/arturo-lang/arturo/pull/1008/commits/2ffb3c1d475eff9d7b24ce6ed00ecd52300827a7): `"lib.files.art"`
> 2. I wrote a tiny bug [here](https://github.com/arturo-lang/arturo/pull/1008/commits/997a1cbbb95edf55412dc3bace91f12af6496907): `write ~"|dir|/test.json" write ~"|dir|/test.json" jsonSample`
> 
> OMG, I think I would run well now. 🙂

I *think* it's now working perfectly.

Only issue is: the `.res` file doesn't correspond to what it should be.

For reference: https://www.diffchecker.com/LLDhQ4pF/ 😉

RickBarretto:
> Only issue is: the .res file doesn't correspond to what it should be.

My bad, I forgot to update it. Since I'd disabled these tests, I'd removed this lines too.

RickBarretto:
😮‍💨
🙂

drkameleon:
@RickBarretto That's awesome! Really! :)

Great job!

If you agree and give me the green light, I think we could probably merge the PR as-is, and we can get back to it when the rest of the issues are fixed. 😉

drkameleon:
Great job one more time @RickBarretto! 😄 

Ready to merge! 🚀

PR #1023, opened 2023-03-03
WIP - [Collections/in?] add new `.deep` option

RickBarretto:
# Description
While `in?` searches for a value only in the first depth block, `.deep` makes search into internal `:blocks`, recursively.

Fixes #958

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [ ] This change requires a documentation update

PR #1045, opened 2023-03-04
Update tests for collections

RickBarretto:
Since my PR #1023 gave me a lot of overhead with output updates, I'm changing how unittests should work.
Comparing outputs doesn't give me what exactly is broken and I must search manually to find. And further, output shows false-positives with any change with whitespaces.

[reference](https://github.com/arturo-lang/arturo/pull/1023#issuecomment-1454844340)

## Type of change

- [ ] Update unittests


drkameleon:
Very nice! Looks much cleaner! 😉 

Go for it! And let me know in case you want me to check anything whatsoever! 🚀

RickBarretto:
I found two functions broken for `:literal` types

image
image

RickBarretto:
New problem found:
image
image
image
image

RickBarretto:
Done! :wink:

RickBarretto:
> `SQLite not available in MINI builds`

Disabling for mini builds...

RickBarretto:
Done! :wink:

RickBarretto:
Note: `sort.by` is running ok on playground, but for some reason, it's not working on my local machine, so I commented its tests for while.

drkameleon:
> I found two functions broken for `:literal` types
> 
> ...
Good catch! 🚀 

This looks very weird! I have no idea what is going on here. I'm opening an issue so that we can keep track of this and solve it! 😉

drkameleon:
> New problems found: image image image image

Hmm... Hmm... 🤔 

`sort` is one of the most interesting built-in methods and one of the most complicated ones to get right.

Basically, my initial intention was to replicate one specific Ruby gem: https://github.com/twitter/twitter-cldr-rb (If you have a quick look into its capabilities, you'll get my point).

Now, that being said, this package is actually based on CLDR... meaning: a vastly complicated, ultra-huge piece of code that works beautifully, but is... well... huge.

So, my goal is that Arturo manages to - at least - offer a decent subset of the aforementioned features, but without all this bulk. Let's see... 😄 

Adding issues for this too! Great catch again! 🚀

drkameleon (approved):
Extremely clean job! Awesome! 👍 

P.S. You have actually created some sort of unit-test library/convention for Arturo, and I must say I love the way it looks *and* works! 🚀

drkameleon:
> Done! 😉

Ready to merge! 🚀

RickBarretto:
Thank you, @drkameleon !!!

RickBarretto:
> P.S. You have actually created some sort of unit-test library/convention for Arturo, and I must say I love the way it looks and works! 🚀

Did you like how it works?

I think, it's the best way to get example of use.
The output is on the same file, and ensure gives the right line of the error.

Also, the next test just can run, if the latest is working, so it avoids a test break another.

Futher, it solves my problem with white spaces and colored output redirect from Msys2.

drkameleon:
> > P.S. You have actually created some sort of unit-test library/convention for Arturo, and I must say I love the way it looks and works! 🚀
> 
> Did you like how it works?
> 
> I think, it's the best way to get example of use considering the output is on the same file.
> 
> For us, ensure gives the right line of the error, the next test just can run, if the latest is working, so it avoids a test break another.
> 
> Futher, it solves my problem with white spaces and colored output redirect from Msys2.

I totally do like it! :)

Using an `ensure` makes perfect sense (most unit-test mechanism work in a similar fashion; it's just that I don't use it too much... lol)

Nim, for instance, does it *all* the time (with their `runnableExamples` trick): https://github.com/nim-lang/Nim/blob/version-1-6/lib/impure/re.nim#L451-L452

All in all, the result is *very* clean, easy to reason about and... basically I love it! Well done! 😉

PR #1057, opened 2023-03-08
[VM/values/comparison] add `>` & `<` support for Date values

### LINKED ISSUE #1039: [VM/values/comparison] add `<` support for Date values
[VM/values/comparison] add `<` support for Date values

currently, `=` is support but not `<` and `>`!

https://github.com/arturo-lang/arturo/blob/ed0410b81b479dcf090b239e2193a45ecf65cd5d/src/vm/values/comparison.nim#L169

```text
                return false

# TODO(VM/values/comparison) add `<` support for Date values
#  currently, `=` is support but not `<` and `>`!
#  labels: critical,bug,values
proc `<`*(x: Value, y: Value): bool {.inline.}=
    if x.kind in {Integer, Floating, Rational} and y.kind in {Integer, Floating, Rational}:
        if x.kind==Integer:
```

2dea89ee3e5666b8e12f43e6fa7e6655b17a750e

### LINKED ISSUE #1040: [VM/values/comparison] add `>` support for Date values
[VM/values/comparison] add `>` support for Date values

currently, `=` is support but not `<` and `>`!

https://github.com/arturo-lang/arturo/blob/ed0410b81b479dcf090b239e2193a45ecf65cd5d/src/vm/values/comparison.nim#L258

```text
                return false

# TODO(VM/values/comparison) add `>` support for Date values
#  currently, `=` is support but not `<` and `>`!
#  labels: critical,bug,values
proc `>`*(x: Value, y: Value): bool {.inline.}=
    if x.kind in {Integer, Floating, Rational} and y.kind in {Integer, Floating, Rational}:
        if x.kind==Integer:
ndex 6c951cc8e4..d5eaa2cdb4 100644
++ b/src/vm/values/custom/vrange.nim
```

5935f3b34a34f454d4eeb38cc7f56bb79cd1a63f

RickBarretto:
# Description

Fixes #1039 #1040

## Type of change

- [x] Update unit-tests

### DIFF
--- a/src/vm/values/comparison.nim
+++ b/src/vm/values/comparison.nim
@@ -12,7 +12,7 @@
 # Libraries
 #=======================================
 
-import lenientops, tables, unicode
+import lenientops, tables, unicode, times
 
 when defined(WEB):
     import std/jsbigints
@@ -253,7 +253,7 @@ proc `<`*(x: Value, y: Value): bool {.inline.}=
                 else:
                     return false
             of Date:
-                return x.eobj < y.eobj
+                return x.eobj[] < y.eobj[]
             else:
                 return false
 
@@ -344,7 +344,7 @@ proc `>`*(x: Value, y: Value): bool {.inline.}=
                 else:
                     return false
             of Date:
-                return x.eobj > y.eobj
+                return x.eobj[] > y.eobj[]
             else:
                 return false
 
--- a/src/vm/values/comparison.nim
+++ b/src/vm/values/comparison.nim
@@ -162,7 +162,7 @@ proc `==`*(x: Value, y: Value): bool {.inline, enforceNoRaises.}=
                     if x.dbKind==SqliteDatabase: return cast[ByteAddress](x.sqlitedb) == cast[ByteAddress](y.sqlitedb)
                     #elif x.dbKind==MysqlDatabase: return cast[ByteAddress](x.mysqldb) == cast[ByteAddress](y.mysqldb)
             of Date:
-                return x.eobj == y.eobj
+                return x.eobj[] == y.eobj[]
             else:
                 return false
 
--- a/src/vm/values/comparison.nim
+++ b/src/vm/values/comparison.nim
@@ -12,7 +12,7 @@
 # Libraries
 #=======================================
 
-import lenientops, tables, unicode, times
+import lenientops, tables, times, unicode
 
 when defined(WEB):
     import std/jsbigints
--- a/src/vm/values/comparison.nim
+++ b/src/vm/values/comparison.nim
@@ -252,6 +252,8 @@ proc `<`*(x: Value, y: Value): bool {.inline.}=
                     return x.proto.doCompare(x, y) == -1
                 else:
                     return false
+            of Date:
+                return x.eobj < y.eobj
             else:
                 return false
 
@@ -341,6 +343,8 @@ proc `>`*(x: Value, y: Value): bool {.inline.}=
                     return x.proto.doCompare(x,y) == 1
                 else:
                     return false
+            of Date:
+                return x.eobj > y.eobj
             else:
                 return false
 
--- a/tests/unittests/lib.comparison.art
+++ b/tests/unittests/lib.comparison.art
@@ -58,6 +58,11 @@ print [1 2 3]=[4 5 6]
 print [1 2 3 4]>[1 2 3]
 
 a: to :date .format: "YYYY" "2002"
+b: to :date .format: "YYYY" "2002"
+print a = b
+print b = a
+
+a: to :date .format: "YYYY" "2002"
 b: to :date .format: "YYYY" "2001"
 print a > b
 print a < b
--- a/tests/unittests/lib.comparison.art
+++ b/tests/unittests/lib.comparison.art
@@ -56,3 +56,18 @@ print [1 2 3]=[1 2 3]
 print [1 2 3]=[4 5 6]
 
 print [1 2 3 4]>[1 2 3]
+
+a: to :date .format: "YYYY" "2002"
+b: to :date .format: "YYYY" "2001"
+print a > b
+print a < b
+
+a: to :date .format: "MMM dd" "Jan 15"
+b: to :date .format: "MMM dd" "Jan 14"
+print a > b
+print a < b
+
+a: to :date .format: "HH:mm" "13:25"
+b: to :date .format: "HH:mm" "12:12"
+print a > b
+print a < b
--- a/tests/unittests/lib.comparison.res
+++ b/tests/unittests/lib.comparison.res
@@ -36,3 +36,9 @@ true
 true
 false
 true
+true
+false
+true
+false
+true
+false
--- a/tests/unittests/lib.comparison.res
+++ b/tests/unittests/lib.comparison.res
@@ -37,6 +37,8 @@ true
 false
 true
 true
+true
+true
 false
 true
 false


RickBarretto:
Surprise, is it reversed?

```
false
true
false
true
false
true
```

But, look out my code:

```nim
172  proc `<`*(x: Value, y: Value): bool {.inline.}=
     ...
255            of Date:
256                return x.eobj < y.eobj
     ...
263  proc `>`*(x: Value, y: Value): bool {.inline.}=
     ...
346            of Date:
347                return x.eobj > y.eobj
      
```

RickBarretto:
@drkameleon, I'll wait for your advise here.
I would invert them, but I don't know if it's the best approach.

src/vm/values/comparison.nim
```diff
@@ -252,6 +252,8 @@ proc `<`*(x: Value, y: Value): bool {.inline.}=
                     return x.proto.doCompare(x, y) == -1
                 else:
                     return false
+            of Date:
+                return x.eobj < y.eobj
```
drkameleon, line 256:
```suggestion
                return x.eobj[] < y.eobj[]
```

src/vm/values/comparison.nim
```diff
@@ -341,6 +343,8 @@ proc `>`*(x: Value, y: Value): bool {.inline.}=
                     return x.proto.doCompare(x,y) == 1
                 else:
                     return false
+            of Date:
+                return x.eobj > y.eobj
```
drkameleon, line 347:
```suggestion
                return x.eobj[] > y.eobj[]
```

drkameleon:
Another sidenote explaining what is happening above:

Since `x.eobj < y.eobj`, etc was actually comparing pointers and e.g. the `2001` date was created *after* the `2002` most likely it was assigned to a "higher" memory address... hence the comparison seems to be yielding the results sort-of reversed. But not really... 😉

RickBarretto:
Sure!

RickBarretto:
With this change, I got this error:
```nim
C:\msys64\home\arturo\src\vm\values\comparison.nim(256, 33) Error: type mismatch: got <DateTime, DateTime>
but expected one of:
func `<`(a, b: VVersion): bool
  first type mismatch at position: 1
  required type for a: VVersion
  but expression 'x.eobj[]' is of type: DateTime
func `<`(x, y: VRationalObj): bool
  first type mismatch at position: 1
  required type for x: VRationalObj
  but expression 'x.eobj[]' is of type: DateTime
func `<`(x: Int; y: int | culong | Int): bool
  first type mismatch at position: 1
  required type for x: Int
  but expression 'x.eobj[]' is of type: DateTime
func `<`(x: int | culong; y: Int): bool
  first type mismatch at position: 1
  required type for x: int or culong
  but expression 'x.eobj[]' is of type: DateTime
func `<`[I: SomeInteger; F: SomeFloat](f: F; i: I): bool
  first type mismatch at position: 1
  required type for f: F: SomeFloat
  but expression 'x.eobj[]' is of type: DateTime
func `<`[I: SomeInteger; F: SomeFloat](i: I; f: F): bool
  first type mismatch at position: 1
  required type for i: I: SomeInteger
  but expression 'x.eobj[]' is of type: DateTime
proc `<`(x, y: bool): bool
  first type mismatch at position: 1
  required type for x: bool
  but expression 'x.eobj[]' is of type: DateTime
proc `<`(x, y: char): bool
  first type mismatch at position: 1
  required type for x: char
  but expression 'x.eobj[]' is of type: DateTime
proc `<`(x, y: float): bool
  first type mismatch at position: 1
  required type for x: float
  but expression 'x.eobj[]' is of type: DateTime
proc `<`(x, y: float32): bool
  first type mismatch at position: 1
  required type for x: float32
  but expression 'x.eobj[]' is of type: DateTime
proc `<`(x, y: int): bool
  first type mismatch at position: 1
  required type for x: int
  but expression 'x.eobj[]' is of type: DateTime
proc `<`(x, y: int16): bool
  first type mismatch at position: 1
  required type for x: int16
  but expression 'x.eobj[]' is of type: DateTime
proc `<`(x, y: int32): bool
  first type mismatch at position: 1
  required type for x: int32
  but expression 'x.eobj[]' is of type: DateTime
proc `<`(x, y: int64): bool
  first type mismatch at position: 1
  required type for x: int64
  but expression 'x.eobj[]' is of type: DateTime
proc `<`(x, y: int8): bool
  first type mismatch at position: 1
  required type for x: int8
  but expression 'x.eobj[]' is of type: DateTime
proc `<`(x, y: pointer): bool
  first type mismatch at position: 1
  required type for x: pointer
  but expression 'x.eobj[]' is of type: DateTime
proc `<`(x, y: string): bool
  first type mismatch at position: 1
  required type for x: string
  but expression 'x.eobj[]' is of type: DateTime
proc `<`(x, y: uint): bool
  first type mismatch at position: 1
  required type for x: uint
  but expression 'x.eobj[]' is of type: DateTime
proc `<`(x, y: uint16): bool
  first type mismatch at position: 1
  required type for x: uint16
  but expression 'x.eobj[]' is of type: DateTime
proc `<`(x, y: uint32): bool
  first type mismatch at position: 1
  required type for x: uint32
  but expression 'x.eobj[]' is of type: DateTime
proc `<`(x, y: uint64): bool
  first type mismatch at position: 1
  required type for x: uint64
  but expression 'x.eobj[]' is of type: DateTime
proc `<`(x, y: uint8): bool
  first type mismatch at position: 1
  required type for x: uint8
  but expression 'x.eobj[]' is of type: DateTime
proc `<`(x: Value; y: Value): bool
  first type mismatch at position: 1
  required type for x: Value
  but expression 'x.eobj[]' is of type: DateTime
proc `<`[Enum: enum](x, y: Enum): bool
  first type mismatch at position: 1
  required type for x: Enum: enum
  but expression 'x.eobj[]' is of type: DateTime
proc `<`[T: tuple](x, y: T): bool
  first type mismatch at position: 1
  required type for x: T: tuple
  but expression 'x.eobj[]' is of type: DateTime
proc `<`[T](x, y: ptr T): bool
  first type mismatch at position: 1
  required type for x: ptr T
  but expression 'x.eobj[]' is of type: DateTime
proc `<`[T](x, y: ref T): bool
  first type mismatch at position: 1
  required type for x: ref T
  but expression 'x.eobj[]' is of type: DateTime
proc `<`[T](x, y: set[T]): bool
  first type mismatch at position: 1
  required type for x: set[T]
  but expression 'x.eobj[]' is of type: DateTime

expression: x.eobj[] < y.eobj[]
  CRASHED!!!
```

RickBarretto:
I'll have a look for help in Nim's discord.

RickBarretto:
I think, I found my error here.
`<` and `>` comparators for `DateTime` comes from [times module](https://nim-lang.org/docs/times.html#DateTime). :wink:

RickBarretto:
Also, actually, Arturo is comparing pointers for `equal? ; =` operations with `:date`.

RickBarretto:
Done! 😉

drkameleon:
> I think, I found my error here. `<` and `>` comparators for `DateTime` comes from [times module](https://nim-lang.org/docs/times.html#DateTime). 😉

I was about to write this until I noticed you've already solved it! Well done! 😉 

**Speaking of that, one word of notice regarding Nim:**

Nim normally works by overloading operators, including the most common ones (`==`, `<`, etc) for the main types. And, these operators, are used under the hood by many, *many* other methods.

So, when you're doing `x in y` (membership test) you're actually using `==`. In other cases you may be needing the `hash` function of a value, etc.

What I'm trying to say here is that, even though it may not always be 100% obvious, we must always make sure that the overloaded version of the underlying operators *are* visible (= imported). Otherwise, even a seemingly inoffensive `if x in y` may end up comparing pointers or... anything for that matter.

Tricky, but manageable if we are cautious. (This one was an example of how confusing it may get... Basically, you got no error, the program compiles fine, but silently the behavior is far different than what we intended... so... )

drkameleon:
As a sidenote of what I'm writing above, a similar struggle of mine with a perfectly valid overloaded method that I simply forgot to make "public" (aka `*`)... in a few words, the correct overload was in place, I *did* import the file, but since the method wasn't visible, it wasn't used... instead, the result was totally different:

https://forum.nim-lang.org/t/9540#62669

drkameleon (approved):

drkameleon:
Ready to merge! 🚀

RickBarretto:
> Tricky, but manageable if we are cautious. (This one was an example of how confusing it may get... Basically, you got no error, the program compiles fine, but silently the behavior is far different than what we intended... so... )

> As a sidenote of what I'm writing above, a similar struggle of mine with a perfectly valid overloaded method that I simply forgot to make "public" (aka *)... in a few words, the correct overload was in place, I did import the file, but since the method wasn't visible, it wasn't used... instead, the result was totally different:
https://forum.nim-lang.org/t/9540#62669


I see! I'll take more caution with it.

Also, I always have problem with Nim's LSP. It's slow to give me the answer of what I'm working, and then I gave up and search manually into codebase.
I takes easily much more than 30s to show me what type is some variable.

drkameleon:
Well, regarding what you're saying, my main issue has been this `nimsuggest` thing. It seems like it's spawning 10 instances of the same process in the background, so much so that at the end I'm left out of RAM and have to manually kill them!

I think I have read somewhere about this... 🤔

PR #1059, opened 2023-03-10
WIP - [Draft] - Cleanup & Rewrite `Collections\sort`

### LINKED ISSUE #1050: [Collections/sort] clean rewrite needed
[Collections/sort] clean rewrite needed

the whole implementation looks like a patchwork of ideas and is not that

easy to debug.

also, there seem to be different types of issues: https://github.com/arturo-lang/arturo/pull/1045#issuecomment-1458960243

https://github.com/arturo-lang/arturo/blob/dd0282eb8b8548b1044a203b3bc6380914df7383/src/library/Collections.nim#L1652

```text
                    push(newBlock())

    # TODO(Collections/sort) clean rewrite needed
    #  the whole implementation looks like a patchwork of ideas and is not that
    #  easy to debug.
    #  also, there seem to be different types of issues: https://github.com/arturo-lang/arturo/pull/1045#issuecomment-1458960243
    #  labels: library, cleanup

    # TODO(Collection/sort) make sure all options work as expected for Literal values too
    #  see: https://github.com/arturo-lang/arturo/pull/1045#issuecomment-1458960243
    #  labels: library, bug, critical
    builtin "sort",
        alias       = unaliased,
        op          = opNop,
```

535ef7f2aa54f8a4cf69a93c96fb1ff8b948b6db

RickBarretto:
# Explanation
It's one of the most difficult functions to debug in Arturo, since it gets multiple types, a lot of attributes and repeats the same code again and again...

## The Structure
- helpers/arrays.nim
  - `sortBlock*`: the function to be called, also with `var` version
  - `SortParams*`: exported params to be used to create a tuple of parameters
  - `sortBlockBy`: a helper function used in case of `.by` attribute
- helpers/dictionaries.nim
  - `sortDictionary*`: the function to be called, also with `var` version
  - `all` (custom): a helper function used to iterate Dictionaries
  - `SortParams`: defines the parameters used by `sortDictionary*` function

### Reducing the error possibility
Reduce everything to two overloaded function (`sortBlock` & `sortDictionary`) avoids errors by miss something here or there.
Even for In-placed values, this process is simplified, since you call the same function, but overloaded with a `var` type.

### Reducing indentation
My approach to reduce indentation was to put attributes inside a tuple and pass it to the functions, instead of `if`s for `checkAttr(...)` and `hadAttr(...)`, then call these functions to get things done.

### Reducing the amount of parameters by call
To do it, I placed every attribute inside a common tuple, to pass to functions. Note that for Dictionary I pass an extra `values` separately.

Tries to fix #1050

## Type of change

- [ ] Code cleanup


RickBarretto:
Side Note: It's just a draft. It's not complete, don't solve any other problem and also, may be broken in some part...

RickBarretto:
Alright! Let's see it! 😝🙂

drkameleon:
> Since we can use `||` now to compare types, can I update and implement this PR?

You mean inside the implementation of `sort`? If it is so, I think you can go ahead without any issue! 😉

drkameleon:
> I'll close this to implement into other PR.

Sure! That sounds like a fine idea :)

drkameleon:
By the way, before going ahead with this, please have a look at this first: https://github.com/arturo-lang/arturo/issues/1178 (it's an issue that has been bothering me for some time, but which could result in some functionality - namely `.ascii` - not being there at all in the first place 😉 )

PR #1068, opened 2023-03-11
[System\sys] Add `hostname` info

### LINKED ISSUE #1062: [Sockets] add function to return current HostName
[Sockets] add function to return current HostName

an even better implementation would be to actually return a dictionary, with different

details about the host, e.g.: `host\name`, etc

use case: https://rosettacode.org/wiki/Hostname

for reference: https://nim-lang.org/docs/nativesockets.html#getHostname

https://github.com/arturo-lang/arturo/blob/3f408e2ed98260b1972f8bd984af826bcb7a3174/src/library/Sockets.nim#L43

```text
#  labels: open discussion

# TODO(Sockets) add function to return current HostName
#  any even better implementation would be to actually return a dictionary, with different
#  details about the host, e.g.: `host\name`, etc
#  use case: https://rosettacode.org/wiki/Hostname
#  for reference: https://nim-lang.org/docs/nativesockets.html#getHostname

proc defineSymbols*() =
    
    when not defined(WEB):
ndex b14eee24d6..20d1c55fcb 100644
++ b/src/library/Strings.nim
```

86723dd599d9f1d1061ac939542f2ae82bd61783

drkameleon:
# Description

Fixes #1062 

## Type of change


drkameleon:
Ready to merge! 🚀

PR #1070, opened 2023-03-13
Sanitize arch names in binaries & releases + workflows cleanup

### LINKED ISSUE #1035: [System/sys] normalize the way CPU architecture is shown
[System/sys] normalize the way CPU architecture is shown

in our new release builds, we annotate x86_64/amd64 builds as "x86_64"

here, our sys\cpu field would return "amd64"

obviously, we should normalize this, but we need to decide on a single name

and then, we need to make sure that all our build scripts are using the same name

https://github.com/arturo-lang/arturo/blob/ed0410b81b479dcf090b239e2193a45ecf65cd5d/src/library/System.nim#L329

```text
                push newLogical(isAdmin())

    # TODO(System/sys) normalize the way CPU architecture is shown
    #  in our new release builds, we annotate x86_64/amd64 builds as "x86_64"
    #  here, our sys\cpu field would return "amd64"
    #
    #  obviously, we should normalize this, but we need to decide on a single name
    #  and then, we need to make sure that all our build scripts are using the same name

    # TODO(System/sys) add info about endianess?
    #  we could add another field to the dictionary, like "endianess" : "big" or "little"
    #  or, perhaps even better, make our existing cpu field a dictionary with two fields:
    #  - type: amd64
    #  - endian: little
    #  ...
    builtin "sys",
        alias       = unaliased, 
        op          = opNop,
ndex 2f1d8eda81..ebb9ce37f7 100644
++ b/src/vm/ast.nim
```

ee1afd16c0d81bbd1cd431d0b78ba17eb98b0863

drkameleon:
# Description

What this PR does:

- Roughly speaking, whenever we had `x86_64` for 64-bit builds, we'll now have `amd64` (same for `sys\cpu\arch` and the binary names)
- If running on a 32-bit system, instead of Nim's Linux-ish `i386`, `sys\cpu\arch` will show `x86` (same as our binary releases)
- Add OpenBSD & NetBSD to the list of supported OS'es (in terms of builds & releases)

Fixes #1035 

## Type of change

- [x] Code cleanup

### DIFF
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -31,7 +31,7 @@ jobs:
         version:
           - mini
         arch:
-          - x86_64
+          - amd64
     steps:
       - name: "Checkout"
         uses: actions/checkout@main
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -34,7 +34,7 @@ jobs:
           - safe
         arch:
           - x86
-          - x86_64
+          - amd64
           - arm
           - arm64
         exclude:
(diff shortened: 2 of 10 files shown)


drkameleon:
Ready to merge! 🚀

PR #1071, opened 2023-03-13
[Core/ensure] add option for custom message

### LINKED ISSUE #1061: [Core/ensure] add option for custom message
[Core/ensure] add option for custom message

so that when `ensure` fails, the message can be customized

use case: https://rosettacode.org/wiki/Assertions_in_design_by_contract

https://github.com/arturo-lang/arturo/blob/3f408e2ed98260b1972f8bd984af826bcb7a3174/src/library/Core.nim#L415

```text
                execUnscoped(x)
            
    # TODO(Core/ensure) add option for custom message
    #  so that when `ensure` fails, the message can be customized
    #  use case: https://rosettacode.org/wiki/Assertions_in_design_by_contract
    builtin "ensure",
        alias       = unaliased, 
        op          = opNop,
ndex 9dae3ddd55..30545bf48a 100644
++ b/src/library/Sockets.nim
```

fa8d759d8cdd972769531588cbecb3d29c68df47

RickBarretto:
# Description

Adds an attribute to print a custom message when `ensure` fails.

Fixes #1061 

## Type of change

- [x] Update/create tests
- [x] Update/create usage example 

### DIFF
--- a/src/vm/errors.nim
+++ b/src/vm/errors.nim
@@ -221,7 +221,8 @@ proc AssertionError_AssertionFailed*(context: string) =
           
 proc AssertionError_AssertionFailed*(context: string, message: string) =
     panic AssertionError, 
-          "\"" & message & "\": " & context
+          message & ":;" & 
+          "for: " & context
 
 # Runtime errors
 
--- a/src/vm/errors.nim
+++ b/src/vm/errors.nim
@@ -218,6 +218,10 @@ proc SyntaxError_EmptyLiteral*(lineno: int, context: string) =
 proc AssertionError_AssertionFailed*(context: string) =
     panic AssertionError,
           context
+          
+proc AssertionError_AssertionFailed*(context: string, message: string) =
+    panic AssertionError, 
+          "\"" & message & "\": " & context
 
 # Runtime errors
 
--- a/src/library/Core.nim
+++ b/src/library/Core.nim
@@ -412,10 +412,7 @@ proc defineSymbols*() =
             if isFalse(y): 
                 execUnscoped(x)
             
-    # TODO(Core/ensure) add option for custom message
-    #  so that when `ensure` fails, the message can be customized
-    #  use case: https://rosettacode.org/wiki/Assertions_in_design_by_contract
-    #  labels: library, enhancement
+            
     builtin "ensure",
         alias       = unaliased, 
         op          = opNop,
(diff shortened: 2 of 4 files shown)


drkameleon:
Very good job! I generally like it. 😄 

**A couple of points:**

- The one is about naming. Wouldn't it be better if it's something a bit shorter than `.message`? I mean... the option is longer than the command itself. I was thinking something more like `.with` (which is used quite extensively) or `.that` (in the sense of `ensure that...` - thinking in English. On the other hand, `.message` is more than self-explanatory. I don't know; I don't have a very concrete suggestion. So, I'm leaving this up to you! 😉 

- The other one is about styling the error message (not that I'm 100% satisfied with the way error message appear anyway, but I'll fix it little by little). Regarding this, my only point is to make it look like the other errors, with the context being shown in a different line, that's all; I'll make an edit suggestion to see what I'm talking about exactly 😉 

Other than the above... beautifully done! 

Also: removing the relevant TODO - which you did - is good too, since the bot won't be confused when merging the PR! 🚀

src/vm/errors.nim
```diff
@@ -218,6 +218,10 @@ proc SyntaxError_EmptyLiteral*(lineno: int, context: string) =
 proc AssertionError_AssertionFailed*(context: string) =
     panic AssertionError,
           context
+          
+proc AssertionError_AssertionFailed*(context: string, message: string) =
+    panic AssertionError, 
+          "\"" & message & "\": " & context
```
drkameleon, line 224:
```suggestion
          message & ":;" & 
          "for: " & context
```

`;`s are a hack to denote *line-change* within an error body. I think - just think - it will hopefully look cleaner and more uniform this way... Also, I don't really think we need the double quotes, IMHO. 😉

RickBarretto:
> in the sense of `ensure that...` - thinking in English.

Following the guidelines of semantics, I think that `that` fits better. 
```red
; Ensure that variable should be positive: ...
ensure.that: "variable should be positive" -> a >= 0
```

RickBarretto:
If you think it's ok, ready to merge! :rocket:

drkameleon:
I love it! Short and looking nice in the end! Good job! 😉

drkameleon (approved):

drkameleon:
Ready to merge! 🚀

PR #1079, opened 2023-03-14
[Converters\to] update usage examples for `:complex` and `:rational`

### LINKED ISSUE #1075: [Converters\to] getting an undefined behavior when creating `:complex` or `:rational` with some negative field
Undefined behavior when creating a `:complex` or a `:rational` with a negative field

image
image

**Desktop (please complete the following information)**
- OS: amd64/windows
- Version v/0.9.83 b/170

**Additional context**
Discovered on #1072

RickBarretto:
# Description

This update explains why you **must** use `@` for negative fields when converting to `:rational`s and `:complex`es, and add missing example for `:rational` type. 🙂 

Related to #1075

## Type of change
- [x] Update documentation

### DIFF
--- a/src/library/Converters.nim
+++ b/src/library/Converters.nim
@@ -1352,6 +1352,8 @@ proc defineSymbols*() =
             to :floating 4                ; 4.0
 
             to :complex [1 2]             ; 1.0+2.0i
+            
+            ; make sure you're using the `array` (`@`) converter here, since `neg` must be evaluated first
             to :complex @[2.3 neg 4.5]    ; 2.3-4.5i
             
             to :boolean 0                 ; false
--- a/src/library/Converters.nim
+++ b/src/library/Converters.nim
@@ -1356,6 +1356,9 @@ proc defineSymbols*() =
             ; make sure you're using the `array` (`@`) converter here, since `neg` must be evaluated first
             to :complex @[2.3 neg 4.5]    ; 2.3-4.5i
             
+            to :rational [1 2]            ; 1/2
+            to :rational @[neg 3 5]       ; -3/5
+            
             to :boolean 0                 ; false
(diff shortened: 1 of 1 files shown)


RickBarretto:
Done! 🚀

drkameleon:
The truth is it's even more generic and part of the power of Arturo's syntax (Rebol/Red share the exact same logic):
the distinction between what is a data type and code is very blurred.

For example, let's take this:

```red
blk: [add 3 2] 
```
if you think it's a block that adds 3 and 2, you're technically wrong ❗ . Basically all we have is a block with 3 things: a `:word` (add), and two `:integer` values (3 & 2).

if you `print do blk` then it will actually evaluate the block and you will see 5 as the result.

But let's do something else instead.

```red
blk: [add 2 3] ; as before

; now, let's... change the contents of the block, as we would with any other block :)
blk\0: to :word "mul"

; well, we just changed the first word from "add" to "mul"
; so if we do the following:
print do blk

; guess what, we've just changed the code programmatically
; and this thing prints... 6!
```
Screenshot 2023-03-15 at 10 45 39

If you think about it, this is *very* powerful. And, of course, I should thoroughly explain this in the documentation, as a general rule (since, as a matter of fact, it has nothing to do with `neg` or `to :complex`, etc per se).

But, in any case, adding some more illustrative examples doesn't hurt! 😉

drkameleon (approved):

drkameleon:
Ready to merge! 🚀

RickBarretto:
> Basically all we have is a block with 3 things: a :word (add), and two :integer values (3 & 2).

Damn, thanks for the explanation. It's much cleaner now!

> But, in any case, adding some more illustrative examples doesn't hurt! 😉

I agree! 😂😂

PR #1080, opened 2023-03-14
[Converters\as] update formatting to support negative fields for `:rational` and `:complex` types

RickBarretto:
# Description
`as.code` wasn't returning valid code for `:rational` and `:complex` types when one of these fields were a negative number.

What I did, was support these cases.

Before:

```red
a: to :rational [3 5]
print as.code a - 1
; => to :rational [-2 5]

b: to :complex [1 2]
as.code b - 2
; => to :complex [-1.0 2.0]
```

Now:
```red
a: to :rational [3 5]
print as.code a - 1
; => to :rational @[neg 2 5]

b: to :complex [1 2]
as.code b - 2
; => to :complex @[neg 1.0 2.0]
```

As we may know, `to rational [-2 5]` is invalid, since `-` is the same as `sub`, and it doesn't turn `2` into a negative number.
So, I decided to turn these numbers into positive, add a `neg` function to it and add a `@` that is needed here.

Related to #1075 

image

## Type of change

### DIFF
--- a/src/vm/values/printable.nim
+++ b/src/vm/values/printable.nim
@@ -474,9 +474,7 @@ proc codify*(v: Value, pretty = false, unwrapped = false, level: int=0, isLast: 
             else:
                 result &= fmt("to :complex [{v.z.re} {v.z.im}]")
         of Rational     : 
-            if v.rat.num < 0 and v.rat.den < 0:
-                result &= fmt("to :rational @[neg {v.rat.num * -1} neg {v.rat.den * -1}]")
-            elif v.rat.num < 0:
+            if v.rat.num < 0:
                 result &= fmt("to :rational @[neg {v.rat.num * -1} {v.rat.den}]")
             else:
                 result &= fmt("to :rational [{v.rat.num} {v.rat.den}]")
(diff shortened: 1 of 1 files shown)


RickBarretto:
Done! 🚀

RickBarretto:
**Note about the `little improvement on performance`**

I reduced the amount of control-flow, since `:rational` automatically converts negative fields to equivalent when possible.
- If only the denominator is negative, then `:rational` reverses the signal, making the numerator negative instead
- If both fields (numerator and denominator) are negative, so `:rational` reverses the signal, making both positive

image

RickBarretto:
Aaaannd, Done! 🙂

drkameleon:
I was a bit skeptical at first I have to admit (I thought it had to do with the actual values of `:complex` or `:rational` values), but nope, it's actually perfect: the problem was with `as.code` (or better put with `codify`). Well done! 🚀 

As a sidenote: `as.code` is one that I wouldn't be based too much for debugging/testing purposes as - I imagine - there are more than a few case where it could still fail. I would mostly go with `inspect` which has been more than ultra-used and tested 😉

drkameleon:
Ready to merge! 🚀

PR #1100, opened 2023-03-22
[Collections\last] improve function's performance for range types

### LINKED ISSUE #1097: [Collections/first & last] better handling for infinite ranges
**Describe the bug**

Right now, trying to get the `last` element of an infinite range (e.g. `1..∞`) provokes a crash.

Also see: https://discord.com/channels/765519132186640445/829324913097048065/1086600060395081860

**Expected behavior**

In that case, we should be returning `infinity`.
In the case of `first (neg ∞)..10`, we should be returning *negative* infinity instead.

RickBarretto:
# Description

There is two problems here, the main problem is described in the issue #1097, and the other is described [in my comment](https://github.com/arturo-lang/arturo/issues/1097#issuecomment-1480090093).

1. `last` was internally trying to generate an infinite block to get the last element. Obviously it crashes, since we don't have infinite memory and it would iterate infinitely until **never** reaches the infinite number... (fixed on [9dd1577](https://github.com/arturo-lang/arturo/pull/1100/commits/9dd15775f69b8fc2d3e16a73cd3279a1108f3657))
2. even solving the 1st problem, it still generates a block (`O(n/step)` amortized) to get the last number, but **we just want the last number**.

Fixes #1097

## TODO

- [x] add case for infinite ranges
- [x] add case for last element
  - [x] cover chars
- [x] add case for the last n elements, returning a new range
  - [x] cover chars

## Type of change
- [x] Bug fix (~non~ breaking change which fixes an issue)
  - returns a `:range` instead of `:block` for `last.n`

### DIFF
--- a/src/helpers/ranges.nim
+++ b/src/helpers/ranges.nim
@@ -105,6 +105,44 @@ func `[]`*(rng: VRange, idx: HSlice): ValueArray =
         )
         start += step
         i += 1
+        
+        
+func `[]`*(rng: VRange, idx: int, returnValue: bool): Value =
+    
+    if idx > rng.len or idx < 0: 
+        raise newException(ValueError, "Index out of range!!")
+
+    let step = 
+        if rng.forward: rng.step
+        else: -1 * rng.step
+    
+    return 
+        if rng.numeric: newInteger(rng.start + step*(idx - 1))
+        else: newChar(char(rng.start + step*(idx - 1)))
+    
+    
+        
+func `[]`*(rng: VRange, idx: HSlice, returnValue: bool): VRange =
+    
+    if idx.a > rng.len or idx.b > rng.len or idx.a < 0 or idx.b < 0: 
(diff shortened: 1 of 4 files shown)


RickBarretto:
I don't know if you'll like my idea of returning a new range, but here is how it's going on my local builds:
image
image


Also, I can make it return a `:block` instead as well, your decision. :100:

drkameleon (approved):
Awesome idea + very clean implementation! 🚀 

Also, regarding your idea about returning a `:range` for `last.n:`, I totally agree. Spot on!

----

P.S. Just a sidenote - I'm not telling you to change anything at all - just so that you know: `to :block rng` obviously works; perhaps, the easiest way to convert a `:range` to its evaluated `:block` would be `@`, like:

```red
print @1..9 
```

😉

RickBarretto:
> the easiest way to convert a :range to its evaluated :block would be @, like:

Ohhhh, thanks for the tip!

PR #1101, opened 2023-03-23
WIP: Sanitize Value kind checking in built-in functions

drkameleon:
# Description

Up to this point, in all built-in function we've been doing various kind checks (e.g. `if x.kind == ...`, etc) and in some cases multiple times. Since - as of the parent PR - the kinds of the passed values (`x`, `y`, `z`) are automatially *pre*-stored (as `xKind`, `yKind` & `zKind` respectively) via `require`, the whole library has to be updated to use these instead of all the `x.kind` mess.

The performance boost may be tiny (I consider this a micro-optimization that will avoid more than 1 lookup for the kind field), but actually it will make the code a tiny bit cleaner as well.

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #1077, opened 2023-03-14
Clean up & rewrite Value operations

drkameleon:
# Description

This PR is meant to perform a thorough cleanup/rewrite of the whole operation-related code for Arturo's main Value objects and in the meantime handle any pending issues along with the appropriate error-reporting.

### Main goals

- cleaner code, without 3 zillion repetitions (preferrably using templates/macros when appropriate)
- fix issues
- better performance
- failures (e.g. non-compatible operations for given Value pairs) have to become runtime errors

### Core methods updated

**Related to Arithmetic module:**

  - [x] [add](https://arturo-lang.io/documentation/library/arithmetic/add)
     - [x] in-place
  - [x] [dec](https://arturo-lang.io/documentation/library/arithmetic/dec)
     - [x] in-place
  - [x] [div](https://arturo-lang.io/documentation/library/arithmetic/div)
     - [x] in-place
  - [x] [divmod](https://arturo-lang.io/documentation/library/arithmetic/divmod)
     - [x] in-place
  - [x] [fdiv](https://arturo-lang.io/documentation/library/arithmetic/fdiv)
     - [x] in-place
  - [x] [inc](https://arturo-lang.io/documentation/library/arithmetic/inc)
     - [x] in-place
  - [x] [mod](https://arturo-lang.io/documentation/library/arithmetic/mod)
     - [x] in-place
  - [x] [mul](https://arturo-lang.io/documentation/library/arithmetic/mul)
     - [x] in-place
  - [x] [neg](https://arturo-lang.io/documentation/library/arithmetic/neg)
     - [x] in-place
  - [x] [pow](https://arturo-lang.io/documentation/library/arithmetic/pow)
     - [x] in-place
  - [x] [sub](https://arturo-lang.io/documentation/library/arithmetic/sub)
     - [x] in-place

**Related to Bitwise module:**

  - [x] [and](https://arturo-lang.io/documentation/library/bitwise/and)
     - [x] in-place
  - [x] [not](https://arturo-lang.io/documentation/library/bitwise/not)
     - [x] in-place
  - [x] [or](https://arturo-lang.io/documentation/library/bitwise/or)
     - [x] in-place
  - [x] [shl](https://arturo-lang.io/documentation/library/bitwise/shl)
     - [x] in-place
  - [x] [shr](https://arturo-lang.io/documentation/library/bitwise/shr)
     - [x] in-place
  - [x] [xor](https://arturo-lang.io/documentation/library/bitwise/xor)
     - [x] in-place

[...]

PR #1104, opened 2023-03-24
Revisit Arithmetic module unit-tests

drkameleon:
# Description

After the rather huge [parent PR](https://github.com/arturo-lang/arturo/pull/1077), it's time to make sure everything works here as it should! 🚀 

@RickBarretto I would appreciate some input here, as this PR progresses... 😉 

### Bugs spotted

- [x] **`add`**: Complex/Rational 
- [x] **`div`**: Complex/Complex (in-place)

## Type of change

- [x] Code cleanup


drkameleon:
Since there are tests in place that check for `null`, and as the code is right now, no operations are returning `null` anymore... I guess we should omit these tests... OR - which is something that I've been thinking:

link the `ensure` with some sort of `try? [] else []` (if this `try?` return `false`, practically it means that it crashed (= threw an exception), which - in the case of combinations that are *not* valid - would be the equivalent of a successful outcome... Hmm... 🤔 

What do you think?

RickBarretto:
>  I would appreciate some input here, as this PR progresses...

Sure! I can help you! :wink:

> pair the ensure with some sort of try? [] else [] (if this try? return false, practically it means that it crashed (= threw an exception), which - in the case of combinations that are not valid - would be the equivalent of a successful outcome... Hmm... 🤔
> What do you think?

This is my initial approach to this, and the right one, IMHO. 
Even I have been using it for some unit-tests that returns an exception.

Some examples:
- [Testing divide by 0 oprations](https://github.com/arturo-lang/arturo/blob/master/tests/unittests/lib.arithmetic.art#L633-L637) and [for floating types](https://github.com/arturo-lang/arturo/blob/master/tests/unittests/lib.arithmetic.art#L669-L679)
- [Testing try to chop a function instead of a variable](https://github.com/arturo-lang/arturo/blob/master/tests/unittests/lib.collections.art#L271-L272)

drkameleon:
> > I would appreciate some input here, as this PR progresses...
> 
> Sure! I can help you! 😉
> 
> > pair the ensure with some sort of try? [] else [] (if this try? return false, practically it means that it crashed (= threw an exception), which - in the case of combinations that are not valid - would be the equivalent of a successful outcome... Hmm... 🤔
> > What do you think?
> 
> This is my initial approach to this, and the right one, IMHO. Even I have been using it for some unit-tests that returns an exception.
> 
> Some examples:
> 
> * [Testing divide by 0 oprations](https://github.com/arturo-lang/arturo/blob/master/tests/unittests/lib.arithmetic.art#L633-L637) and [for floating types](https://github.com/arturo-lang/arturo/blob/master/tests/unittests/lib.arithmetic.art#L669-L679)
> * [Testing try to chop a function instead of a variable](https://github.com/arturo-lang/arturo/blob/master/tests/unittests/lib.collections.art#L271-L272)
> 
> I'm current using this pattern:
> 
> ```red
> try? -> itWillFail
> else -> passed
> ```

Awesome! 🚀  I guess I kind-of missed it!

My main goal with this PR is to merely make it pass, by spotting things that I may have gotten wrong re-writing the value operations and un-commenting the ones that are currently working.

Other than that, when it's finished, you are more than welcome to just change it as you please, or revise it, or add more tests (I guess there's still some room!). At least, which value operations are permitted is currently more than obvious, just by looking at the implementation! 😉

drkameleon:
Btw, @RickBarretto re-thinking about...

Yes, `try?`/`else` are mostly used together, with `else` being triggered when the preceding `try?` "failed". But... I'm thinking. Basically, `try?` as all predicates (`...?`) do, actually returns a Logical value (true/false), in the case of `try?` returning whether the block (inside `try?`) ran successfully.

So, wouldn't just using `if not try? [] -> passed` work?

RickBarretto:
> So, wouldn't just using if not? try? itWillFail-> passed work? (or it's shorter equivalent: unless try? itWillFail -> passed

Yeah, for sure! If you prefer this pattern, we can use too. 🙂
If you want, I can update every already implemented unit-tests to this new pattern.
I liked this new approach, much more readable, I think.

drkameleon:
@RickBarretto I'm going through the unit-test and trying to spot issues.

So I'm looking at this:

```
ensure -> null = a + b
ensure -> null = b + a
ensure -> null = add a b
ensure -> null = add b a
passed
```

Since, right now, none of this operations returns `null` (they throw an error instead), I thought of using `try?` as we were saying.

So, I thought of this:

```
ensure -> not? try? -> a + b
ensure -> not? try? -> b + a
ensure -> not? try? -> add a b
ensure -> not? try? -> add b a
```

or...

```
unless try? -> a + b -> passed
unless try? -> b + a -> passed
unless try? -> add a b -> passed
unless try? -> add b a -> passed
```

(ofc, in that case, we'll have 4 times "passed")

And then I thought of doing it in a tiny bit more idiomatic way:

```
ensure -> not any? @[
     try? -> a + b
     try? -> b + a
     try? -> add a b
     try? -> add b a
]
passed
```

(Basically, `try?` will return true only if the block - e.g. `a+b` - was executed without an error. `any?` checks if *any* of the given expressions in the block returns *true*. But we obviously don't expected any of the expressions to return true - that is: to be executed without throwing an error), so... we `ensure -> not?...`)

I think it looks quite easy on the eye, nope? Plus, I think it follows your way of doing things, which is already *very* clean! :)

RickBarretto:
Yeaah, for sure!
I loved the lastest one! Easy to write, read and clean!

I loved this approach, avoiding repetitive tasks...

drkameleon:
Me: I'll quickly fix the failing tests and move on with the PR. Shouldn't take more than 10 minutes!

🕐 🕑 🕒 

Me (3 days later): Just a few more failing tests to fix... 🤣

RickBarretto:
An awesome (and loooong, lol) work! 

A hard work, but with a good benefit! I loved it!

Also, I have two questions:
1. Can I use `||` for any built-in function? 2. Is `X || Y` the same of `Y || X `, or does it return different values?

drkameleon:
> An awesome (and loooong, lol) work! 🚀
> 
> A hard work, but with a good benefit! I loved it!

Thanks! 😄 

> Also, I have two questions:
> 
> 1. Can I use `||` for any built-in function?
> 2. Is `X || Y` the same of `Y || X `, or does it return different values?

Two very good questions.

**(1)** Not right now, but - as you put - I think we *should* be able to use it for any built-in function as well, so I'm thinking of putting it in some global module so that it's available from anywhere. Thanks for reminding me! 😉 

**(2)** The reason for the existence of this new syntax (`||`) is precisely that (with even more extra... sauce). 

Let me explain:
     
Given how Arturo is designed internally and how Nim works, Values are actually object pointers and their *kind* (used merely to create object variants) is actually a C-style [**enum**](https://en.wikipedia.org/wiki/Enumerated_type), that is a set of "named" values.

So, basically, a Value's kind can be an Integer, a String, or [anything from this list](https://github.com/arturo-lang/arturo/blob/master/src/vm/values/types.nim#L63-L101); btw, as you can see, each Value already has a number assigned to it. So, what can be done internally (and Nim is quite good at this) is treat this "types" as numbers.

Let's imagine something...

Let's say we have an array of 64 elements.

|  64 | 63  | 62  | 61  | ...  | 5  | 4  | 3  |  2 | 1  |
|---|---|---|---|---|---|---|---|---|---|


So, what we'll do is: for kind 1 (which is, in our case, Logical), we'll tick the rightmost cell of array table. Let's say this "tick" is a 1 (so, the rest of our cells will be 0's). And so on...

In that way, we could have:
```
Logical (1)  = `00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000001`
Integer (2)  = `00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000010`
Floating (3) = `00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000100`
Complex (4)  = `00000000 00000000 00000000 00000000 00000000 00000000 00000000 00001000`
```

And so on...

Now, if you rethink about it, what we're actually doing is making use of the CPUs architecture... and since `int`s (in 64-bit system) are to be assumed to have 64-bits, what we are actually doing is correlate our Value kinds with actual *int* values (Logical ends up being 1, Integer is 2, Floating is 4, Complex is 8, etc - just calculate the values above as binary values 😉 )

So... we've ended up with a slightly-bizarre but very efficient way of "codifying" our Value kinds.

Now, how would we refer to 2 value kinds at the same time?

Let's say: express "Complex or Integer".

Let's take our code from above:

```
Integer (2)  = `00000000 00000000 00000000 00000000 00000000 00000000 00000000 00000010`
Complex (4)  = `00000000 00000000 00000000 00000000 00000000 00000000 00000000 00001000`
```

if we "add" the bits (= that's a binary OR), then we'd get:

```
IntegerOrComplex  = `00000000 00000000 00000000 00000000 00000000 00000000 00000000 00001010`
```

which is actually a binary number, with 2 bits set, one at the "Integer" position and another one at the "Complex" position.

Now let's say we have a Value's kind (let's call this Unknown) as a number. (created as above ^)

How would we check if it's an "Integer or complex"? All we'd have to do is:

```Unknown AND IntegerOrComplex = Unknown```

Basically, we check for the common bits and if there is a coincidence, it's a match.

Now, the best question would be: why go through all this fuss? Because AND/OR operations on simple numbers is *extremely* fast and the CPU loves this type of things... 😉 

------

Now, back to the original question:

Can this type of encoding 2 values (as the IntegerOrComplex example) "see" the order of the values? The answer would be: no.

And that's why I re-designed the whole value checking process, so that `X || Y` is *not* the same as `Y || X` (because in our use case we actually *do* care about the order...; if we didn't, and we were to check for one value, something like `x.kind in {Integer,Floating}` would work flawlessly, with Nim internally transforming it to something along the lines of what I described above ^ 😄  )

Also, the `X || Y` syntax can also identify out-of-the-box if an Integer is a BigInteger (which is rather common and *very* important, even though it's not even a Value kind, properly speaking - it's just a kind for the Value objects that are Integer's, so it's even more complicated 😉 )

And another long, long-winded explanation has come to an end... 🤣

RickBarretto:
>  I think we should be able to use it for any built-in function as well,

🙂 I'm happy to hear it!

> Because AND/OR operations on simple numbers is extremely fast and the CPU loves this type of things... 😉

Really loved it! Woow!

> And another long, long-winded explanation has come to an end... 🤣

I really love your explanations. 🙂

PR #1105, opened 2023-03-24
[Collections\pop] creating a new `pop` function

### LINKED ISSUE #1024: [Collections] add a proper `pop` built-in method
[Collections] add a proper `pop` built-in method

right now, `chop` works like that, kind-of, but does *not* return the

removed item

also: Core/pop actually pop an item from the VM stack, which *is* useful,

but a) not so common, b) we might need a different name for that

https://github.com/arturo-lang/arturo/blob/ed0410b81b479dcf090b239e2193a45ecf65cd5d/src/library/Collections.nim#L49

```text
#=======================================

# TODO(Collections) add a proper `pop` built-in method
#  right now, `chop` works like that, kind-of, but does *not* return the
#  removed item
#
#  also: Core/pop actually pop an item from the VM stack, which *is* useful,
#  but a) not so common, b) we might need a different name for that
#

proc defineSymbols*() =

    builtin "append",
ndex d8944eb50c..9db11e37a1 100644
++ b/src/library/Converters.nim
```

aae6e9290d546ebb8f28be0e44ce5cb9f7199aee

RickBarretto:
# Description

This code adds a proper Collections\`pop` function, similar to Collections\`chop`, but returning the popped value(s).
Also, I removed the old Core\`pop`, that will be updated into another PR.

Fixes #1024
- Related PR #1106

## Type of change

- [x] Breaking change
  - remove old Core\`pop`


RickBarretto:
Done!

RickBarretto:
@drkameleon, done!
This PR is related to #1106, and it can have some conflicts, since I removed Core\`pop`.
Also, this purely won't works, since `console.art` depends on Core\`pop` that I had to remove here. But this behavior is fixed on #1106. You can do what you want.

But I recommend to merge #1106 first and then merge it...

drkameleon (approved):
Pure awesomeness! 😉 

Clean implementation, with examples, types all set fine, everything looking good!! 🚀 

----

P.S. As a tiny side-comment: I wouldn't be using `as.code` too much; personally, I use it only for actually unwrapping code text. In most case, a simple`print` (at least for example code) would do. As for actually testing a value, I tend to use `inspect`, which is more than tried and tested and produces results in a rather visually-ok way 😉

drkameleon:
Ready to merge! 🚀  

❗ after #1106 is merged

RickBarretto:
All right! I'll update this later! :wink:

RickBarretto:
Done! 🚀

PR #1107, opened 2023-03-26
[Collections\first] improve performance for range types, as it was for [Collections\last]

RickBarretto:
# Description

As `first` works as `last`, we should update it too. So, I did it! 
Now, it returns a `:range` instead of a `:block` while using `first.n: :integer :range`, and also as it doesn't create a `:block` anymore, it can be considered a performance improvement for large ranges...

- Related with #1100 

## Type of change

- [x] Add/Update unit-tests


RickBarretto:
Done! :rocket:

drkameleon (approved):
Awesome job, one more time! 😄

PR #1111, opened 2023-03-28
[Core] added new `throws?` method

### LINKED ISSUE #1109: Core: Add new `throws?` function
The function will behave as a light-weight alias for `not? try?` (which doesn't look intuitive at all), and will basically return *true* if a block of code has thrown an exception, or *false* otherwise. The `.verbose` option (of `try?`) is *not* needed - since we merely want to see if a block actually "throws" (en exception) or not; that's all 😉 

The context might seem a bit unnecessary. but:

- The method could come *very* handy for us when used e.g. in unit-tests (the existing lib.arithmetic unit-tests for instance is already flooded with `not? try?`s, that could be vastly simplified this way
- The function is to be used in conjunction with other Exception-related methods that have already been planned; namely: 
   - #1026
   - #1028
   - and... ultimately #1043)

😉

drkameleon:
# Description

Fixes #1109 

## Type of change


drkameleon:
Ready to merge! 🚀

Issue #1121, opened 2023-03-29
Add proper, more self-explanatory errors + type-checking for Block params


drkameleon:
### Overview

Right now, for standard library functions that take a Block, we perform the usual type-checking (that it's a Block) and that's it.

Now, what if this Block's element have to be of a specific kind? 

Let's take `sum` for example... it takes a Block as a parameter. But could it take a Block of e.g. String values? The answer is obviously *no*. Could it takes a mixed Block? Nope. So in that case, we should trigger an error for the first non-compatible Value found.

In other cases, we may need to perform this type of type-checking even before doing any processing and show the appropriate message beforehand.

And there are other cases, where the block we expect has even stricter characteristics: e.g. `to :complex [...]` takes 2 arguments (not more, not less!) that have to be either Integer's or Floating's. So, that's another one.

**Summing up:**

- we have to create *one*, uniform way of checking for this type of scenarios that will be available throughout the standard library
- we have to - at least try - show all errors, where appropriate

----

**Also see:** https://github.com/arturo-lang/arturo/pull/1117#pullrequestreview-1362557047

drkameleon:
Regarding this scenario:

> Let's take sum for example... it takes a Block as a parameter. But could it take a Block of e.g. String values? The answer is obviously no. Could it takes a mixed Block? Nope. So in that case, we should trigger an error for the first non-compatible Value found.

This is a rather special case, which I'm very tempted to leave as-is right now. (Btw, `sum` & `product` work almost in identical way)

So, how does `sum` work right now?

<img width="343" alt="Screenshot 2023-04-04 at 13 03 26" src="https://user-images.githubusercontent.com/1265028/229772609-c984f02e-62e3-4a44-8d88-0820a379c855.png">

Which may initially seem strange, but is it? Actually `sum` is implemented as a sequential addition of the given block's elements. And addition triggers this type of errors. So, partly it would be normal to expect coming from `sum` directly, but:

- I think the error is understable enough
- Most other interpreted languages would behave in a similar way, showing similar errors (= related to the actual operation being performed; and failing)

Hence, in order not to add extra performance hurdles, I think the wisest choice would be to leave it exactly as it is right now.

PR #1124, opened 2023-03-29
WIP - [Numbers\clamp] use Range instead of `min`/`max` as parameter

### LINKED ISSUE #1122: [Numbers/clamp] use a Range instead of min/max, as parameters
[Numbers/clamp] use a Range instead of min/max, as parameters

given how elegantly Range values are used, it would be better to use e.g a `range` 

Range type parameter for this one instead of a `max` and `min` parameter, so that we

we could do something along the lines of `clamp 2 1..3` (would return 2) or

`clamp 4 1..3` (would return 3) or `clamp 0 1..3` (would return 1); same as before.

Also, should we reverse the order of the params, and have the range as the first one?

https://github.com/arturo-lang/arturo/blob/aa86b48a2cc12d7349ed57a7879e08135ab0e870/src/library/Numbers.nim#L395

```nim
            push(newInteger(int(ceil(asFloat(x)))))

    # TODO(Numbers/clamp) use a Range instead of min/max, as parameters
    #  given how elegantly Range values are used, it would be better to use e.g a `range` 
    #  Range type parameter for this one instead of a `max` and `min` parameter, so that we
    #  we could do something along the lines of `clamp 2 1..3` (would return 2) or
    #  `clamp 4 1..3` (would return 3) or `clamp 0 1..3` (would return 1); same as before.
    #  Also, should we reverse the order of the params, and have the range as the first one?
    builtin "clamp",
        alias       = unaliased, 
        op          = opNop,
```

f431b59c616f16350f111c2a0b3defd7ad1108fa

RickBarretto:
# Description

Changes:
- [x] `clamp` accepts a range instead of two parameters.
  - [x] should work with reversed range
  - [x] should work for range with step different of 1
  - [ ] Should work with `:char`s
  - [ ] Should not work for `:floating` as first parameter, since `:range`s doesn't accept

- Fixes #1122

## Type of change

- [x] Add/Update unit-tests

### DIFF
--- a/src/library/Numbers.nim
+++ b/src/library/Numbers.nim
@@ -393,13 +393,7 @@ proc defineSymbols*() =
             #=======================================================
             push(newInteger(int(ceil(asFloat(x)))))
 
-    # TODO(Numbers/clamp) use a Range instead of min/max, as parameters
-    #  given how elegantly Range values are used, it would be better to use e.g a `range` 
-    #  Range type parameter for this one instead of a `max` and `min` parameter, so that we
-    #  we could do something along the lines of `clamp 2 1..3` (would return 2) or 
-    #  `clamp 4 1..3` (would return 3) or `clamp 0 1..3` (would return 1); same as before.
-    #  Also, should we reverse the order of the params, and have the range as the first one?
-    #  labels: library, enhancement
+
     builtin "clamp",
         alias       = unaliased, 
         op          = opNop,
--- a/src/library/Numbers.nim
+++ b/src/library/Numbers.nim
@@ -29,6 +29,7 @@ when not defined(NOGMP):
 
 import helpers/maths
 import helpers/ranges
+import vm/values/custom/vrange
 
 import vm/errors
 
@@ -406,8 +407,7 @@ proc defineSymbols*() =
         description = "force value within given range",
         args        = {
             "number" : {Integer,Floating},
-            "min"    : {Integer,Floating},
-            "max"    : {Integer,Floating}
+            "range"  : {Range}
         },
         attrs       = NoAttrs,
         returns     = {Integer,Floating},
@@ -417,12 +417,15 @@ proc defineSymbols*() =
             clamp 4 1 3             ; 3
         """:
             #=======================================================
-            if x < y:
-                push(y)
-            elif x > z:
-                push(z)
+            if y.rng.forward:
+                if x.i < y.rng.start: push(newInteger(y.rng.start))
+                elif x.i > y.rng[y.rng.len, true].i: push(newInteger(y.rng[y.rng.len, true].i))
+                else: push(x)
             else:
-                push(x)
+                if x.i > y.rng.start: push(newInteger(y.rng.start))
+                elif x.i < y.rng[y.rng.len, true].i: push(newInteger(y.rng[y.rng.len, true].i))
+                else: push(x)
+            
 
     builtin "conj",
         alias       = unaliased, 
--- a/src/library/Numbers.nim
+++ b/src/library/Numbers.nim
@@ -412,9 +412,11 @@ proc defineSymbols*() =
         attrs       = NoAttrs,
         returns     = {Integer,Floating},
         example     = """
-            clamp 2 1 3             ; 2
-            clamp 0 1 3             ; 1
-            clamp 4 1 3             ; 3
+            clamp 2 1..3                ; 2
+            clamp 0 1..3                ; 1
+            clamp 4 1..3                ; 3
+            clamp 4 3..1                ; 3
+            clamp 5 range.step: 2 1 5   ; 4
         """:
             #=======================================================
             if y.rng.forward:
--- a/tests/unittests/lib.numbers.art
+++ b/tests/unittests/lib.numbers.art
@@ -1,0 +1,31 @@
+topic: $[topic :string] -> print ~"\n>> |topic|"
+passed: $[] -> print "[+] passed!"
+
+topic « clamp
+do [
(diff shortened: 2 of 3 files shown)


RickBarretto:
Done! 🚀 
But, some notes:
- We could support `:char` too, but it's unnecessary and would be strange, since it is on `Numbers` lib.
- Instead of create a new `RuntimeError_*` function, I opted for use `panic RuntimeError, ...` inline, I don't know if you see it as a problem

src/library/Numbers.nim
```diff
@@ -392,37 +393,48 @@ proc defineSymbols*() =
...
+            
+            if x.kind == Integer:
+                if y.rng.forward:
+                    if x.i < y.rng.start: push(newInteger(y.rng.start))
+                    elif x.i > y.rng[y.rng.len, true].i: push(newInteger(y.rng[y.rng.len, true].i))
+                    else: push(x)
+                else:
+                    if x.i > y.rng.start: push(newInteger(y.rng.start))
+                    elif x.i < y.rng[y.rng.len, true].i: push(newInteger(y.rng[y.rng.len, true].i))
+                    else: push(x)
```
drkameleon, line 427: Very nice job! 🚀 

Also, I like the handling of `.step:` (basically, I totally agree with your approach) and considering char ranges as invalid; it would be just too confusing... 😉 

```suggestion
                if (let minElem = y.rng.min()[1]; x.i < minElem.i): push(minElem)
                elif (let maxElem = y.rng.max()[1]; x.i > maxElem.i): push(maxElem)
                else: push(x)
```

Since we already have the [`min`](https://github.com/arturo-lang/arturo/blob/master/src/helpers/ranges.nim#L150-L158) / [`max`](https://github.com/arturo-lang/arturo/blob/master/src/helpers/ranges.nim#L160-L168) helper methods defined for Range values, and they do precisely that (return the min/max Value of a Range, regardless of whether it's a forward one or not; along with the index of that Value), why not just this? 😉 

Also, we can avoid getting the last/first elements twice, by using Nim's version of let-if syntax above. 😉  (Not the most intuitive way to write this sort of things, but what can we do...)

src/library/Numbers.nim
```diff
@@ -392,37 +393,48 @@ proc defineSymbols*() =
...
             else:
-                push(x)
+                if y.rng.forward:
+                    if x.f < y.rng.start.toFloat: push(newInteger(y.rng.start))
+                    elif x.f > y.rng[y.rng.len, true].i.toFloat: push(newInteger(y.rng[y.rng.len, true].i))
+                    else: push(x)
+                else:
+                    if x.f > y.rng.start.toFloat: push(newInteger(y.rng.start))
+                    elif x.f < y.rng[y.rng.len, true].i.toFloat: push(newInteger(y.rng[y.rng.len, true].i))
+                    else: push(x)
```
drkameleon, line 436:
```suggestion
                if (let minElem = y.rng.min()[1]; x.f < float(minElem.i)): push(minElem)
                elif (let maxElem = y.rng.max()[1]; x.f > float(maxElem.i)): push(maxElem)
                else: push(x)
```

Same as above ^

src/library/Numbers.nim
```diff
@@ -392,37 +393,48 @@ proc defineSymbols*() =
...
+        returns     = {Integer, Floating},
         example     = """
-            clamp 2 1 3             ; 2
-            clamp 0 1 3             ; 1
-            clamp 4 1 3             ; 3
+            clamp 2 1..3                ; 2
+            clamp 0 1..3                ; 1
+            clamp 4 1..3                ; 3
+            clamp 4 3..1                ; 3
+            clamp 5 range.step: 2 1 5   ; 4
```
drkameleon, line 413:
```suggestion
            clamp 5 range.step: 2 1 5   ; 4
```

This is exactly the part that made me test the code myself... but I think it's just a typo. It should returns (*and* returns) 5, no? 😄

drkameleon (commented):
Also... regarding errors, I get your point (I'm not pro-bloating the code without a valid reason), but:

since Arturo is in the process of having a tiny bit more elaborate error reporting system AND I've been [thinking of making the type checking e.g. for Block values even better](https://github.com/arturo-lang/arturo/issues/1121), I think some initial "Incompatible X error, expect Y" would make sense.

Plus, I think it would be ready for me to use for that scenario! 🚀 

So, I would suggest something like that:

```nim
proc RuntimeError_IncompatibleValueType*(functionName: string, tp: string, expected: string) =
    panic RuntimeError,
          "cannot perform _" & (functionName) & "_;" &
          "incompatible value type for " & tp & ";" &
          "expected: " & expected
```

which you could then call like:

```nim
            if not y.rng.numeric:
                RuntimeError_IncompatibleValueType("clamp", valueKind(y), "numeric range")
```

I think it doesn't look bad at all IMHO:

Screenshot 2023-03-30 at 11 25 44

😄

src/library/Numbers.nim
RickBarretto: Yes, it was a mistyping, my bad. The range should be in this case: `range.step: 2 0 5`, since it generates the following array: `[0 2 4]`.

src/library/Numbers.nim
RickBarretto: Well, my bad. I had forgot about `min`/`max` functions. LOL!

RickBarretto:
Done! 🙂

drkameleon (approved):
Looking totally fine!

Great job (again!) and fast! 😉

drkameleon:
Ready to merge! 🚀

PR #1134, opened 2023-04-01
WIP: Better error reporting for incompatible values, etc

### LINKED ISSUE #1121: Add proper, more self-explanatory errors + type-checking for Block params

drkameleon:
# Description

Fixes #1121 

## Type of change

- [x] Code cleanup


drkameleon:
### UPDATE

**Wrong type of value within block parameter:**

Screenshot 2023-04-04 at 18 00 48

**Wrong type of value within block attribute:**

Screenshot 2023-04-04 at 18 00 15

drkameleon:
Ready to merge! 🚀

PR #1139, opened 2023-04-04
Add unit-test for Comparison Library

RickBarretto:
# Description

Unit-test for [Comparison Lib](https://arturo-lang.io/documentation/library/comparison/).

This'll be a veeeery long list...

- [x] `between?`
- [x] `comapare`
  - [ ] `:database`
- [x] `equal?`
  - [ ] `:database`
- [x] `greater?`
  - [ ] `:database`
- [x] `greaterOrEqual?`
  - [ ] `:database`
- [x] `less?`
  - [ ] `:database`
- [x] `lessOrEqual?`
  - [ ] `:database`
- [x] `notEqual?`
  - [ ] `:database`
- [x] `same?`
  - [ ] `:database`

## Type of change

- [x] Add/Update unit-tests

### DIFF
--- a/tests/unittests/lib.comparison.art
+++ b/tests/unittests/lib.comparison.art
@@ -4937,17 +4937,18 @@ do [
     
     topic « less? - :word
     
-    ; ensure -> less? [a] [b]
-    a: [a]
-    b: [b]
-    ensure -> less? a\0 b\0
-    ensure -> not? less? [word] [word]
-    ensure -> not? less? [word] [otherWord]
-    ensure -> not? less? [word] ["word"]
-    ensure -> not? less? [word] [word:]
-    ensure -> not? less? [word] ['word]
-    ; ensure -> not? less? [word] [.word]
-    ensure -> not? less? [word] [.word:]
+    ensure -> less? to :word "a" to :word "b"
+    ensure -> not? less? to :word "a" to :word "a"
+    ensure -> not? less? to :word "b" to :word "a"
+    passed
+    
+    ensure -> not? less? to :word "a" "b"
+    ensure -> not? less? to :word "a" `b`
+    ensure -> not? less? to :word "a" 'b
+    ensure -> not? less? to :word "a" to :label "b"
+    ensure -> not? less? to :word "a" to :attribute "b"
+    ensure -> not? less? to :word "a" to :attributeLabel "b"
+    ensure -> not? less? to :word "word" to :type "z"
     passed
     
     
--- a/tests/unittests/lib.comparison.art
+++ b/tests/unittests/lib.comparison.art
@@ -2567,38 +2567,20 @@ do [
     passed
     
     
-    topic « Equal? - :label
+    topic « greater? - :label
     
-    ensure -> not? greater? [a:] [b:]
-    ; ensure -> greater? [b:] [a:]
-    a: [a:]
-    b: [b:]
-    ; ensure -> greater? a\0 b\0
-    ensure -> not? greater? [label: value] [label: value]
+    ensure -> greater? to :label "b" to :label "a"
+    ensure -> not? greater? to :label "a" to :label "a"
+    ensure -> not? greater? to :label "a" to :label "b"
     passed
     
-    ensure -> not? greater? [label: value] [label: value2]
-    ensure -> not? greater? [label: value] [label2: value]
-    passed
-    
-    ensure -> not? greater? [label:] ["label"]
-    ensure -> not? greater? ["label":] ["label"]
-    ensure -> not? greater? [label:] ["value"]
-    ensure -> not? greater? [label:] ["value"]
-    ensure -> not? greater? [label: "value"] ["label" "value"]
-    passed
-    
-    ; ensure -> not? greater? [label:] [label]
-    ensure -> not? greater? [label:] [value]
-    ensure -> not? greater? [label: "value"] [label value]
-    passed
-    
-    ensure -> not? greater? [label:] ['label]
-    ensure -> not? greater? [label: 'world] ['label 'word]
-    passed
-    
-    ; ensure -> not? greater? [label: .attr] [.label .attr]
-    ; ensure -> not? greater? [label: .attr: "a"] [.label: .attr: "a"]
+    ensure -> not? greater? to :label "a" "a"
+    ensure -> not? greater? to :label "a" `a`
+    ensure -> not? greater? to :label "a" 'a
+    ensure -> not? greater? to :label "a" to :word "a"
+    ensure -> not? greater? to :label "a" to :attribute "a"
+    ensure -> not? greater? to :label "a" to :attributeLabel "a"
+    ensure -> not? greater? to :label "word" to :type "word"
     passed
     
     
--- a/tests/unittests/lib.comparison.art
+++ b/tests/unittests/lib.comparison.art
@@ -6111,20 +6111,20 @@ do [
     ensure -> lessOrEqual? "Art" "Artu"
     passed
     
-    ensure -> not? lessOrEqual? "a"    `a` 
-    ensure -> not? lessOrEqual? "10"   10 
-    ensure -> not? lessOrEqual? "10"   10
-    ensure -> not? lessOrEqual? ["help"] [help]
-    ensure -> not? lessOrEqual? ["a"]    [a:]
-    ensure -> not? lessOrEqual? "help"   'help
-    ensure -> not? lessOrEqual? ["help"] [.help]
-    ensure -> not? "a"      =< `a` 
-    ensure -> not? "10"     =< 10 
-    ensure -> not? "10"     =< 10
-    ensure -> not? ["help"] =< [help]
-    ensure -> not? ["a"]    =< [a:]
-    ensure -> not? "help"   =< 'help
-    ensure -> not? ["help"] =< [.help]
+    ensure -> not? lessOrEqual? "a" `a`
+    ensure -> not? lessOrEqual? "a" `b`
+    ensure -> not? lessOrEqual? "a" 'a
+    ensure -> not? lessOrEqual? "a" 'b
+    ensure -> not? lessOrEqual? "a" to :word "a"
+    ensure -> not? lessOrEqual? "a" to :word "b"
+    ensure -> not? lessOrEqual? "a" to :label "a"
+    ensure -> not? lessOrEqual? "a" to :label "b"
+    ensure -> not? lessOrEqual? "a" to :attribute "a"
+    ensure -> not? lessOrEqual? "a" to :attribute "b"
+    ensure -> not? lessOrEqual? "a" to :attributeLabel "a"
+    ensure -> not? lessOrEqual? "a" to :attributeLabel "b"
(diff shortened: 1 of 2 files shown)


RickBarretto:
`equal?` is almost complete remaining just `:store` and `:database` which I don't know how to test them properly.

drkameleon:
Awesome job here! 😉 I've been observing it closely! Keep it up! 🚀 

Re: databases... The truth is databases is are kind of tricky to test in a CI/unit-test context. 

Let me have a look into it!

RickBarretto:
Adding tests for incompatible types...

RickBarretto:
I covered incompatible types too...

RickBarretto:
All right! Now I will review and format the code to see if everything is right!

RickBarretto:
I'll rewrite some tests now...
Look:
image

https://arturo-lang.io/playground/?et1d51#terminal

As we may know, `equal? 1 1` returns `false`, so I should break the `ensure -> not? all?`, so I'll replace it by `every?`. My bad here...

RickBarretto:
All right! Note that in lines `617` and `1689` I'm using this:

```red
617. ensure -> not? every? @[
618.        -> equal? 1 1
619. ] => not?
```

This assertions ensures if this kind of structure is working well, avoiding misbehaviors from the unit-test itself. I'm not even using `passed` at the end, since it's not a test, but a kind of "meta-test" (Let's call it that...).

drkameleon:
I've been looking into your work. Very nicely done and *very* organized. Just awesome! 🚀 🚀 

Re this:

> As we may know, equal? 1 1 returns false

...I'm a tiny bit confused... What do you mean?

RickBarretto:
> ...I'm a tiny bit confused... What do you mean?

My bad! 🤣🤣🤣🤣
I mean, i t returns false to `ensure`, since we are using an `not?`.
`equal? 1 1` actually returns `true` not `false` itself.

Basically, when I was using the `not? all?` version, it was not actually working. Since if we put an `equal? 1 1` (that returns true) won't break the test, what shouldn't happen, since we want only `false` values.

drkameleon:
> > ...I'm a tiny bit confused... What do you mean?
> 
> My bad! 🤣🤣🤣🤣 I mean, i t returns false to `ensure`, since we are using an `not?`. `equal? 1 1` actually returns `true` not `false` itself.
> 
> Basically, when I was using the `not? all?` version, it was not actually working. Since if we put an `equal? 1 1` (that returns true) won't break the test, what shouldn't happen, since we want only `false` values.

Haha! Phewww... ok, I got it!

I mean... ok, I'd expect a bug or two somewhere, but I wouldn't expect something so bad! lol 🤣

drkameleon:
And in the meantime, I decided to mess thing up and closed (and re-opened) the PR.

I mean... I have 2 buttons "Close with comment" - "Comment", I guess it's a test whether I'm paying enough attention! lol

RickBarretto:
>  I guess it's a test whether I'm paying enough attention! lol

I noticed it! And, It happens to me everytime. 🤣🤣🤣
I look to both buttons, wait for 6 sec to finally click the right button.
It's the reason why I usually use `Ctrl` + `Enter` instead.

RickBarretto:
For some reason comparing if temperature are the `same?` returns false:

```red
print same? 1:c 1:c
; false
print same? 1:f 1:f
; false
print same? 1:k 1:k
; false
```

And an additional note, `print 1:k` prints `1º K`, but Kelvin doesn't is written with `º`.

The playground link: http://arturo-lang.io/playground?uU7NBk

drkameleon:
> For some reason comparing if temperature are the `same?` returns false:
> 
> ```red
> print same? 1:c 1:c
> ; false
> print same? 1:f 1:f
> ; false
> print same? 1:k 1:k
> ; false
> ```

This looks like a total bug! Great catch!

> And an additional note, `print 1:k` prints `1º K`, but Kelvin isn't written with `º` in physics.

Noted!

Basically, regarding Quantity values (and units), I've working on it for the past many days, non-stop. But my plan is to totally re-write the whole thing, so that it's a) cleaner, b) without bugs, c) easier to deal with (similar job as with the value operations), d) far more complete.

Not an easy task by any stretch of the imagination... But I'm determined to make it work as it should. If others have, without even the support of a proper, complete language behind it (see: Frink, etc; which I still use for inspiration in the background along with e.g. [Insect](https://github.com/sharkdp/insect)), we can definitely make it work too, and be great at it. So... let's see! 😉

drkameleon:
That's an interesting suggestion (and I'm already kind-of using Nim's `distinct` in different ways).

The problem with this particular challenge is that it's far, *far* more complicated than it initially looks.

I have studied another - very thorough - implementation that already exists, 100% written in Nim: [Unchained](https://github.com/SciNim/Unchained). It's practically written in... macros.

The challenge here is we don't actually need a perfect Unit system in Nim. We want a perfect Unit system in Arturo, with whatever we may need backstage in Nim. And that's quite daunting.

Basically, if we start easy... it's quite obvious to have e.g. `m` -> `cm` and all that. But then, there are tons of prefixes (you know, all this "cm", "mm", "Gm" -> yes, that exists too: a Giga-meter!, etc). And then we end up with compound units. E.g. if we support the notion of `m` and `s` (= second), how do we define something like `m/s`, which could also be `yd/ns` (= yard / nanosecond), and still refer to speed. And ofc, we could have something like: `1 kg*m / s^2` (which looks like gibberish, but it's actually 1 N(ewton), that is a Power unit). 

The whole thing is extremely interesting I have to admit.... Let's see...

I'll keep your suggestion in mind and I might sneak it in at some point. Right now, I'm pretty close to my 3-4th rewrite... To see the progress (and the resulting mess - that I'm re-re-revisiting right now, as we speak basically), have a look at [this](https://github.com/arturo-lang/arturo/blob/aa1f55e401c55802cbc772dd93e518ffb200926c/src/vm/values/custom/vunit.nim). 🎸 

P.S. If I don't end up totally crazy after this, I'll be lucky... lol 🤣

RickBarretto:
> To see the progress (and the resulting mess - that I'm re-re-revisiting right now, as we speak basically), have a look at this. 🎸

I like how it's going on! Very clean, IMO.

drkameleon:
> I like how it's going on! Very clean, IMO.

Imagine that it looks like a total mess to me (the side-effect of working with the exact same code for hours in a row lol).

I'll show you when I start having more... close (= 🎯 ) to what I have in mind. 😉

RickBarretto:
For some reason, this code returns SIGSEGV, just for Windows...

RickBarretto:
For some reason, this code [returns a SIGSEGV for Windows (time: 3m24s)](https://github.com/arturo-lang/arturo/actions/runs/4691047889/jobs/8314990693?pr=1139), and it runs infinitelly for [MacOs (+15m)](https://github.com/arturo-lang/arturo/actions/runs/4691047894/jobs/8314989019?pr=1139) and[ Linux - returning the error 143 -(+15m)](https://github.com/arturo-lang/arturo/actions/runs/4691047880/jobs/8314988887?pr=1139).

RickBarretto:
I tried to return to [update output - `f3f68ef`](https://github.com/arturo-lang/arturo/pull/1139/commits/f3f68ef8e35850b48ce0c0a2265d55fa373d0a26), but it still returning SIGSEGV. And as you can see, the output and code was completely right.

drkameleon:
> I tried to return to [update output - `f3f68ef`](https://github.com/arturo-lang/arturo/pull/1139/commits/f3f68ef8e35850b48ce0c0a2265d55fa373d0a26), but it still returning SIGSEGV. And as you can see, the output and code was completely right.

Not sure what to say, I'll have to look into myself...

At first sight, I would say this `-> equal? 1:c var 'one?` looks... potentially dangerous, let's say. But I may be wrong... 🤔

RickBarretto:
> > I tried to return to [update output - `f3f68ef`](https://github.com/arturo-lang/arturo/pull/1139/commits/f3f68ef8e35850b48ce0c0a2265d55fa373d0a26), but it still returning SIGSEGV. And as you can see, the output and code was completely right.
> 
> Not sure what to say, I'll have to look into myself...
> 
> At first sight, I would say this `-> equal? 1:c var 'one?` looks... potentially dangerous, let's say. But I may be wrong... 🤔

All right, removing this!

RickBarretto:
> `-> equal? 1:c var 'one?`

It's not the error...

RickBarretto:
Annnnd, I found. `-> same? true var 'true?`. But `true?` is a function that doesn't exists...

RickBarretto:
Notes about `compare` (unittests not implemented yet):

```red
compare 1 "1"
; => 1
compare "1" 1
; => 1
compare 1 2
; => -1
compare ["hello"] #[name: "Joe"]
; => 1
compare #[name: "Joe"] ["hello"]
; => 1
```

Should it happen? `compare`ing different types always result into `1`...

drkameleon:
> Notes about `compare` (unittests not implemented yet):
> 
> ```red
> compare 1 "1"
> ; => 1
> compare "1" 1
> ; => 1
> compare 1 2
> ; => -1
> compare ["hello"] #[name: "Joe"]
> ; => 1
> compare #[name: "Joe"] ["hello"]
> ; => 1
> ```
> 
> Should it happen? `compare` different types always result into `1`...

Hmm... This doesn't look very... correct... 🤔 

Basically, `compare` should be working pretty much like a `<`/`=`/`>`, all-in-one equivalent, where in the first case it would return `-1`, in the second (equality) it would return `0` and the last case would return `1`.

I'll look into it. Well spotted! 😉

RickBarretto:
I think that some filter should be used, instead of delimit `:any` as parameter, the types should be chosen one-by-one.
What I mean:

```red
ensure -> between? null null null
```

Why can I use `null` in between? It should not be possible, IMO.

I'm not saying to create a big control-flow and analyze each type, but instead, just adjust type-by-type `args` field, that currently is:

```nim
args = {
    value"     : {Any},
    rangeFrom" : {Any},
    rangeTo"   : {Any}
},
```

RickBarretto:
Other note, between doesn't work when the 2nd parameter is greater than the 3rd one...

```red
between? 1 0 3
; => true
between? 1 3 0
; => false
```

RickBarretto:
@drkameleon, After I run out all the energy drink from the closest supermarket to me, I finally finished it!!! 🚀

drkameleon:
@RickBarretto This is some extreme piece of work!!! Like, seriously... wow! 😲 

Extremely thorough, detailed, methodical, all very well organized...

Please, give me some time to go through it (mostly, so as to be able to open all issues that are to be opened) and it will be merged.

Again, **awesome** job!!! 😉 🚀 

----

P.S. Apparently the energy drinks did work... lol

celtic-coder:
> P.S. Apparently the energy drinks did work... lol
Apparently! 👍 😄 Great work!

RickBarretto:
Thanks guys! @celtic-coder @drkameleon

drkameleon (approved):
Totally approved. Great job again!

We may have ended up with ~17 more issues to solve, but that's exactly the point! 😉

drkameleon:
And... 🥁 🥁 🥁 

...ready to merge! 🚀

PR #1171, opened 2023-04-25
WIP: Use rationals as internal values for quantities

drkameleon:
# Description

The choice of our internal values is quite crucial.

Generally, if an app requires a high degree of precision and accuracy, we should consider using rational numbers to store the value part of the Quantity type. Rational numbers can represent fractions exactly and can avoid rounding errors that occur when using floating-point numbers. 

However, working with rational numbers can be more complex and computationally expensive than working with floating-point numbers.

Let's see if this idea works and/or it's worth the fuss...

## Type of change

- [ ] Code cleanup


drkameleon:
Ready to merge! 🚀 

(Let's merge this one back, so that we have this all in one place; too many PRs one on top of another and I'll drive myself mad... lol)

PR #1174, opened 2023-04-26
WIP: Add literal syntax for Rational values

drkameleon:
# Description

Given that - if the parent PR succeeds - Rational values will be used as the underlying/internal values for Quantity values, having a better syntax for them woulnd't be bad at all. (I guess, it's far more intuitive than e.g. `to :rational [1 2]`)

Now, speaking of intuitive... the most obvious solution would be `1 / 2` or `1/2` or something similar. However, this totally clashes with the `/` symbol and the way the parser works.

So, the idea/experiment here is to see if we can do it with a `:` (colon). Yep, I know that it may seem like a weird choice, but:

- `1:2` is another way of expressing ratios (that is: `1/2` aside), so it's natural by definition
- `1:2` or in whichever other way we write is pretty much meaningless (no, it wouldn't imply assigning 2 to 1 lol, this would simply crash)
- number parsing already accounts e.g. for scientific notation (which already implies a parser lookahead), so the perceived performance hit should be minimal to none

So... time to experiment! 🚀 

## Type of change


drkameleon:
Ready to merge! 🚀 

(Again, the PR is experimental and there *are* pending issues; but they are to be dealt with in the parent PR 😉 )

Issue #1131, opened 2023-03-31
[Converters\to] Revise how `:rational` works for `:floating` parameters?


RickBarretto:
**Describe the bug**

When I try to create a rational with `1.5` and `3` as parameter, it returns me a `0/1` (the same as `1`), but as we know, the result should be `1/2`, look:

```red
$ to :rational [1.5 3]
; => 0/1
$ 1.5 // 3
; => 0.5
```

**Desktop (please complete the following information):**
- version : `v/0.9.83 b/331 @ 2023-03-30T15:09:55-03:00`
- arch    : `amd64/windows`

**Additional context**
- Found while doing #1130

drkameleon:
Ooops! That's a total bug, doesn't look good at all!

Good that you spotted this! 😉 🚀

drkameleon:
Looking into the whole implementation and how different languages handle Rational numbers, I start think that perhaps the best approach would be to disallow `:floating` types altogether. Generally, rational numbers are defined as numbers consisting of an integer numerator and an integer denominator.

Let me investigate a bit more... 😉

RickBarretto:
> I start think that perhaps the best approach would be to disallow :floating types altogether. 

`:rational` is doing simplifications well, so I think we can use floating as input. But the best approach to do `:floating` -> `:rational` should multiply by 10^x and use this number as denominator.

Eg.: `1.5` should be transformed into `15/10` and just pass it to `:rational` that will do the simplification job that is working well.

Real example: `to :rational [1.5 2]`, should send `[15 20]` to `:rational` constructor.

Just an idea, well you know what fits better.

drkameleon:
> > I start think that perhaps the best approach would be to disallow :floating types altogether.
> 
> `:rational` is doing simplifications well, so I think we can use floating as input. But the best approach to do `:floating` -> `:rational` should multiply by 10^x and use this number as denominator.
> 
> Eg.: `1.5` should be transformed into `15/10` and just pass it to `:rational` that will do the simplification job that is working well.
> 
> Real example: `to :rational [1.5 2]`, should send `[15 20]` to `:rational` constructor.
> 
> Just an idea, well you know what fits better.
> 
> But, as you said, `1.5/2` for instance doesn't makes sense at all.

I guess I'll have to brush up my math skills about rational numbers and their `:floating` equivalent! Oh my, oh my... lol

RickBarretto:
[...]

PR #1176, opened 2023-04-27
Cleanup VRational implementation

drkameleon:
# Description

...

## Type of change

- [x] Code cleanup


drkameleon:
@RickBarretto Apparently, I have managed to fix [this](https://github.com/arturo-lang/arturo/issues/1131) as well, exactly as you suggested (and yes, in the end I think you were 100% right). It might not be the most common thing one would do, but on the other hand... "why not?!" 😉 

**How it works as per this PR** (along with some showcase of the new literal-Rational syntax 😄 ):

Screenshot 2023-04-28 at 09 58 12

drkameleon:
Ready to merge! 🚀

RickBarretto:
> @RickBarretto Apparently, I have managed to fix this as well, exactly as you suggested (and yes, in the end I think you were 100% right).

Wow, thanks a lot!

I have some questions, is this a new thing?

```red
1:2 * 3:4
```

It's much better than use `to :rational`, and more, as `:` is already used for other things and there is no specific function to rational (we use `to` instead) I think it's evaluated at parsing time, what means that `[1:3]` is parsed as a `[:rational]` without the need of use `array` function like this `@[1:3]` to recognize the internal value as a `:rational`, I may be wrong because I didn't test it yet, but it looks good to me.

You know, if `:` is an alias to a function it will not be evaluated at parsing-time, and will return:

```red
inspect [3:2]
; [ :block
;     2 :integer
;     : :symbol
;     3 :integer
; ]
```

Anyway, even if `[1:3]` isn't the same as `[:rational]` it's good enough to me too. 🙂

drkameleon:
Also, so that you can appreciate the changes a bit, since Rational numbers in Arturo can now handle infinitely large (or small) ratios, this is also possible:

Screenshot 2023-04-28 at 16 41 03

(In the previous version, you'd get either 0, or something totally random; and... wrong)

PR #1207, opened 2023-06-13
[Quantities] Bulk-fetch currency exchange rates

drkameleon:
# Description

Right now, since Quantity values generally resolve their base units at creation time, currency Quantity values do that too; but with a catch: every currency creation leads to a network request (to fetch the exchange rate to dollars/USD). That is not practical and rather slow.

**For that:**

- we could downoad the whole list of exchange rates (always with USD as the base unit) from here: https://cdn.jsdelivr.net/gh/fawazahmed0/currency-api@1/latest/currencies/usd.min.json
- ~~optionally, store the whole thing as an Arturo `Store` and~~ use that, if we don't need to re-download it

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #1227, opened 2023-06-15
Fix `inspect` for Quantity values

### LINKED ISSUE #1226: [VM/values/printable] Quantity values should be properly `inspect`able
[VM/values/printable] Quantity values should be properly `inspect`able
right now, `inspect x` where x is a Quantity, invokes *inspect* on the VQuantity value
which was pretty much used in order to test the VQuantity implementation, but not meant
to be publicly visible!

https://github.com/arturo-lang/arturo/blob/55e5b66d8103727fe32541105a616cf97906fc63/src/vm/values/printable.nim#L178

```text

proc dump*(v: Value, level: int=0, isLast: bool=false, muted: bool=false, prepend="") {.exportc.} = 
    # TODO(VM/values/printable) Quantity values should be properly `inspect`able
    #  right now, `inspect x` where x is a Quantity, invokes *inspect* on the VQuantity value
    #  which was pretty much used in order to test the VQuantity implementation, but not meant
    #  to be publicly visible!
    # labels: values, cleanup, bug
    if v.kind==Quantity:
        v.q.inspect()
        return
ndex 2fe2fba745..86d94c629d 100644
++ b/src/vm/values/value.nim
```

d2f5624438f102ab021148e32033c6657f1dfb9d

drkameleon:
# Description

Fixes #1226 

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #1228, opened 2023-06-15
[VQuantity] Clean up `convertTemperature`

### LINKED ISSUE #1225: [VQuantity] clean up `convertTemperature`
[VQuantity] clean up `convertTemperature`
right now, `convertTemperature` is called when given quantity *is* a temperature AND
the target units are compatible - hence a temperature again. So, the `else` statements
below should be considered redundant. Thus, we don't have to output anything! (in case we did,
even then, the errors should be properly thrown errors, coming from VM/errors)

https://github.com/arturo-lang/arturo/blob/55e5b66d8103727fe32541105a616cf97906fc63/src/vm/values/custom/vquantity.nim#L322

```text
        return v

    # TODO(VQuantity) clean up `convertTemperature`
    #  right now, `convertTemperature` is called when given quantity *is* a temperature AND
    #  the target units are compatible - hence a temperature again. So, the `else` statements
    #  below should be considered redundant. Thus, we don't have to output anything! (in case we did,
    #  even then, the errors should be properly thrown errors, coming from VM/errors)
    # labels: cleanup, enhancement, values

    if fromU == K_CoreUnit:
        if toU == degC_CoreUnit:
            result = v - 273.15
```

fd45e6bbedaa3f124c1cb58243680f8db5c8ec5d

drkameleon:
# Description

...

Fixes #1225 

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #1230, opened 2023-06-15
[Env] Fix `parseCmdlineArguments`

### LINKED ISSUE #1218: [Env\parseCmdlineArguments] should return empty values array when no arguments a...
[Env\parseCmdlineArguments] should return empty values array when no arguments are given
currently, calling `args` from a script with no arguments returns a values block with the
script name as the first and only "argument" (which it's not)

https://github.com/arturo-lang/arturo/blob/d5a70bab253681cd020723cbd783be6c137970d7/src/vm/env.nim#L76

```text
# TODO(Env\parseCmdlineArguments) verify it's working right
#  labels: vm,library,language,unit-test

# TODO(Env\parseCmdlineArguments) should return empty values array when no arguments are given
#  currently, calling `args` from a script with no arguments returns a values block with the
#  script name as the first and only "argument" (which it's not)
#  labels: vm,library,language,enhancement
proc parseCmdlineArguments*(): ValueDict =
    ## parse command-line arguments and return 
    ## result as a Dictionary value
ndex c1cd1b7d32..074be54093 100644
++ b/src/vm/exec.nim
```

88d71d41baa389b92d1950b9d7997f501078db0e

drkameleon:
# Description

...

Fixes #1218 

## Type of change


drkameleon:
Ready to merge! 🚀

PR #1233, opened 2023-06-16
[helpers\webview] Properly raise error for browser related things

### LINKED ISSUE #1222: [Helpers/webviews] should produce valid error messages
[Helpers/webviews] should produce valid error messages
currently, we are just outputing a string. Preferrable, it should be done
with a proper error being thrown and declared in VM/errors

https://github.com/arturo-lang/arturo/blob/55e5b66d8103727fe32541105a616cf97906fc63/src/helpers/webviews.nim#L135

```text
            break

    # TODO(Helpers/webviews) should produce valid error messages
    #  currently, we are just outputing a string. Preferrable, it should be done
    #  with a proper error being thrown and declared in VM/errors
    #  labels: enhancement, error handling
    if chromePath == "":
        echo "could not find any Chrome-compatible browser installed"
    else:
ndex 4554f82ed9..306b3e5ea5 100644
++ b/src/helpers/xml.nim
```

e6746ae225c75071ecae3ec5025be105ee35a3a5

RickBarretto:
# Description

Add a proper error throw, instead of print it on the terminal.

Summary:
- Creates `RuntimeError_CompatibleBrowserNotFound`: when the compatible browser is not found, it raises an error
- Creates `RuntimeError_CompatibleBrowserCouldNotOpenWindow`: when the compatible browser can't open a window, it raises an error
- Replaces all `echo "some error message"` by these respective runtime errors.

----

- Fixes #1222

## Type of change

### DIFF
--- a/src/helpers/webviews.nim
+++ b/src/helpers/webviews.nim
@@ -137,7 +137,7 @@ proc openChromeWindow*(port: int, flags: seq[string] = @[]) =
     #  with a proper error being thrown and declared in VM/errors
     #  labels: enhancement, error handling
     if chromePath == "":
-        echo "could not find any Chrome-compatible browser installed"
+        RuntimeError_CompatibleBrowserNotFound()
     else:
         let command = chromePath.replace(" ", r"\ ") & " " & args.join(" ")
         if execCmd(command) != 0:
--- a/src/helpers/webviews.nim
+++ b/src/helpers/webviews.nim
@@ -138,7 +138,7 @@ proc openChromeWindow*(port: int, flags: seq[string] = @[]) =
     else:
         let command = chromePath.replace(" ", r"\ ") & " " & args.join(" ")
         if execCmd(command) != 0:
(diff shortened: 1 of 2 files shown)


RickBarretto:
Done! :wink:

drkameleon (commented):
Very nice job! Exactly what I had in mind! :)

Could we do something about this too? https://github.com/arturo-lang/arturo/blob/2c9d662a4d251dadc88b8a05921ca257e42a065d/src/helpers/webviews.nim#L141

The error is not exactly the same, so - I think - what would make sense would be to make the error message a tiny bit more generic (like "could not open any Chrome-compatible browser") and point to the same error.

What do you think?

RickBarretto:
Ohhh, I forgot that!
Good catch and good idea!

RickBarretto:
Done! :wink:

drkameleon:
Ready to merge! 🚀

PR #1140, opened 2023-04-06
WIP: Prepare for Nim v2.0

drkameleon:
# Description

As [Nim v2.0 is getting closer to its official release](https://forum.nim-lang.org/t/10051), I think it would be wise to start fixing things ahead of time, so that we're not caught by surprise.

Also, there are going to be [more than a handful of changes](https://nim-lang.org/blog/2022/12/21/version-20-rc.html), some of the more critical than others, which means that we have to be prepared for that too... 😉 

So, here we go! 🚀 

## Type of change

- [x] Code cleanup
- [x] Bug fix (non-breaking change which fixes an issue)

PR #1254, opened 2023-08-07
Upgrade to Nim 2.0.0

drkameleon:
# Description

This PR will be an attempt to clean up [the whole mess that was created here](https://github.com/arturo-lang/arturo/pull/1140) and hopefully a) update all code so that it works with the latest Nim release while b) everything keeps working, on *all* OS'es.

## Type of change

- [x] Code cleanup


drkameleon:
So... right now, the whole thing compiles fine (and *should* do so in all OS'es).

The problem is: Arturo is not working at all. And I'm not talking about the usual suspect; it's not even working on macOS. 🙄 

Also: I haven't really changed anything. All I did was to update some missing dependencies (e.g. DB-related modules) that went from being included in Nim's stdlib to being external modules. But other than that, we're talking about literally *zero* core changes.

Now, what remains would be to deal with various new warnings... But I doubt it will make such a huge difference.

Let's see... 🤔

drkameleon:
**Related:** (and super-important) https://forum.nim-lang.org/t/10391

drkameleon:
Ready to merge! 🚀

PR #1291, opened 2023-11-06
New package manager implementation

drkameleon:
# Description

~~First of all, let's not panic! That's just an experiment.~~

Generally, I'm rather frustrated with pretty much all package managers I've used, and the reason why I generally avoid "publishing" new packages. The only one I'm pretty familiar with (and - coincidentally? - the only one for which I've ever published any packages) is RubyGems. So, unavoidably I'm prone to be quite influenced by that to some extent

Now, what do I currently have in mind for Arturo? (I'm not a wizard, nor a package-manager guru, so what I currently have in mind may... change lol)

- Something that is and flexible to use
- By "use" I mean both install/use existing packages, or create/publish our own
- It would be very practical and easy for people creating packages too (I think) to base the whole thing off existing repositories: do you have your code in a repository? then it's a valid "package", that should be directly usable via Arturo
- given that each "package" may carry its own details, we could have an extra file, in the root of our code folder (written like an Arturo dictionary-file for example?) with all the information needed in there (for example: entry file, version, author, documentation url, dependencies, etc)
- we should have one simple command (I'd say `import` - clear, to-the-point, and not too ambiguous) that will handle the whole thing and will work for none-packages as well:
    - Let's say we do `import {something}` - the first thing to do would be to look for a `something.art` file in the exact same folder as the file in question and if found, do something equivalent to `do {something.art}` (which currently works)
    - Let's say the file is not found, then we'd look into the system libraries folder - currently `~/.arturo/lib` (if we call them "packages" - in agreement with our own [pkgr.art](https://pkgr.art) domain - this would need renaming); perhaps we could also include a search in some "hidden" subfolder too before? like `./.pkg`? Makes sense...
    - Now, if the package is not found anywhere, it would have to try and connect with our package registry (`pkgr.art`) and see if there is package named like that; is there a package? then, Arturo should *silently* install it - that is put it in our libraries folder - and continue as if nothing happened (next time the app runs, everything will be in place); if the package is not found *anywhere*, obviously that would result in a runtime error
- generally, the `import` command should also support simple repo urls too (as a way to instantly import things that may be publicly available, but not in our package registry). So, something like this should work too: `import {https://github.com/drkameleon/myartpackage}` (Swift does this in a similar-ish fashion and the truth is it works quite nicely). In that case, the whole thing would work as described above, with the package downloaded and installed locally
- Now, why a registry since we can simply import urls? Basically, having a package in the registry will be something like an extra point for a specific package, since all packages will be pre-approved. How? We could do it by have a PR (automatically created) before accepting any new package. Once the PR is approved, the new package goes in. Plus, having a package in the registry will mean that we will be able to `import {mypackage}` - which admittedly looks even better.

### Extra challenges

- I'm not very confident in Arturo 's **path-handling** (I mean, we'll seriously have to look into this and make sure it works in the most predictable way - no reason to re-invent the wheel here. A few examples of what I mean (@RickBarretto you are perfectly free to help out with this if you have time 😉 ):
     - let's say we do `arturo myapp.art` and in `myapp.art`... 
              - we `read "somefile"`, where is it looking for `somefile`?
              - we `do "somefile"`, where is it looking for `somefile`?
              - we `read "subfolder/somefile"`, is it working?
              - we `do "subfolder/somefile"`, is it working?
              - we `read "../somefile"`, is it working?
              - we `do "../somefile"`, is it working?
     - let's say we do `arturo subfolder/myapp.art` and in `myapp.art`...       
              - we `read "somefile"`, where is it looking for `somefile`?
              - we `do "somefile"`, where is it looking for `somefile`?
              - we `read "subfolder/somefile"`, is it working? and how *should* it be working? 
              - we `do "subfolder/somefile"`, is it working? and how *should* it be working? 
              - we `read "../somefile"`, is it working? and how *should* it be working? 
              - we `do "../somefile"`, is it working? and how *should* it be working? 
- Right now, `do {somefile}` opens the file, reads it and executes it. Now, let's say the file in question contains a function. One would normally expect to `do`/`import` the file and have the function available in the very next line. It *should*(?) be true but it doesn't actually work like that. Basically, for reasons that have to do with the parsing/code-generation stages, etc (and I won't bother you again with that, I promise! lol), in order for us to use the imported functions (the problems is only with functions after all!) we'd have to put our main code inside a block, like that:
     ```red
     import {somefile}
     do [
           myFunc "yeah!" ; now we can use it!
     ]
     ```
     Note: we could also use the much more elegant `do ::` syntax and that's it. But the "issue" remains. And until we introduce some weird pre-processing stage to Arturo (which would definitely kill performance, so I'm really questioning if we should in the first place), I guess we'd just have to live with that... Plus, if I were to actually postpone the implementation of some sort of package manager until then, I'm afraid we'd end up waiting until 2046 😛 
- Having the whole thing based on repo's, would that mean we'd have to 100%-depend on the user having **git** installed? Is this even an issue? (Just thinking aloud...)
-----

So, these are just a few(? - not really) thoughts of mine on the subject. I'll keep experimenting - with code. And let's see how it goes. Of course, any input/feedback/ideas are more than welcome! 🚀 

## Also related

- 🏭  https://github.com/arturo-lang/pkgr.art 
- 📦 https://github.com/arturo-lang/dummy-package 

## Fixes

Fixes: #626 
Fixes: #1293 

## Type of change


RickBarretto:
What an awesome news!
Yes I can help you with this.

I'll have a better looking on this tomorrow.

drkameleon:
> What an awesome news! Yes I can help you with this.
> 
> I'll have a better looking on this tomorrow.

I've been working mostly on https://github.com/arturo-lang/pkgr.art (or https://github.com/arturo-lang/pkgr.art/blob/main/tools/getrepodetails.art to be precise), since this will probably include the core part and I can tell you that this feels exciting. And - generally - it seems to be flowing nicely (I have come across a couple of issues with Arturo, so I'll keep publishing issues here on GitHub, but none of them is that much of a dealbreaker...)

Let's see, let's see! :)

drkameleon:
Btw, the inspiration behind the domain name (`pkgr.art`) lies in one simple fact:

- you know you can `do` a string (and Arturo will parsing it and execute it), e.g. `do {print 123}`
- you know you can `do` a block
- you know you can `do` a file (and Arturo will open it, read it, parse it and execute it), e.g. `do "myfile.art"`
- but, do you know you can actually `do` a URL? Try this and you'll realize it works: 😃 

```red
do "https://raw.githubusercontent.com/arturo-lang/arturo/master/examples/projecteuler/euler1.art"
```
Screenshot 2023-11-07 at 07 17 56

Basically, given the inherent flexible in Arturo's strings, we can actually run scripts anywhere in the web, just by passing the URL!.

So, I've been thinking that... given that there *is* an `.art` TLD (pretty much like our own file extension), it wouldn't look bad if somebody could also do something like:

```red
import {mylib.pkgr.art}
```

... would it? 😉 

(of course, as I explain above, the `import` command will not be just a simple `do` - and the `pkgr.art` serve will be doing its magic regarding what it serves (it still won't be a simple text file -  it'll be much more than that... but maintaining some consistency even in that case wouldn't be bad... 🚀 )

drkameleon:
Well, well, well...

The PR has pretty much reached a very functional point, so I guess it's time to wrap the whole thing up.

In a few words...

Ready to merge! 🚀

RickBarretto:
Excited for this, I'm going to test it right now! (●'◡'●)

RickBarretto:
I tried every step that you mentioned in the begining of this PR, here is the final result:
image

PR #1296, opened 2023-11-07
Make Date values codifiable

### LINKED ISSUE #1295: [Values/printable] `as.code` not working for Date values
**Describe the bug**

`as.code x` (with `x` being a Date value) should return a code representation of the value.

**To Reproduce**

```red
as.code now
```

This returns nothing - an empty string!

**Expected behavior**

Something like: `to :date "2023-11-06T18:35:41+01:00"`

where `2023-11-06T18:35:41+01:00` is what `to :string DateValue` would return.

**Additional context**

Basically, the fix should be ultra-simple.

⚠️  In [Values/printable (codify)](https://github.com/arturo-lang/arturo/blob/master/src/vm/values/printable.nim#L465), the Date implementation is missing. It should return `to :date XXXX` where `XXXX` is the string-ified version of a Date value as returned [from here](https://github.com/arturo-lang/arturo/blob/master/src/vm/values/printable.nim#L113). Shouldn't take more than 5 minutes - including compiling the whole thing. 😉

drkameleon:
# Description

Fixes #1295 

## Type of change


drkameleon:
Ready to merge! 🚀

PR #1298, opened 2023-11-11
Cleaner version info

drkameleon:
# Description

Right now, if we open the REPL it shows complete version information: main version + build + system (+ build timestamp).

It wouldn't be bad since the main information (everything but the timestamp, which is not that useful + it creates some compile-time constants issues) should also be shown when doing `arturo --version`.

Also, we have to normalize the way a couple of build details are displayed (e.g. it shows to me "macosx" - while the official name of the OS is "macOS" since 2016...)

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #1299, opened 2023-11-11
Disable non-working terminal options

drkameleon:
# Description

Some options are - currently - simply not working. E.g. `--update` (that was a daring one) and `--module` (which is [currently been re-implemented](https://github.com/arturo-lang/arturo/pull/1291) from the ground up).

So, instead of having them around totally half-baked, let's just disable them altogether and re-include them when the time comes! 

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #1304, opened 2023-11-18
Experimental ~:exception~ `:error` type implementation

### LINKED ISSUE #1043: [Vm/values/types] add new `:exception` type?
[Vm/values/types] add new `:exception` type?

this could work well with a new potential `try?`/`catch` syntax, 

or a `throw` method

https://github.com/arturo-lang/arturo/blob/ed0410b81b479dcf090b239e2193a45ecf65cd5d/src/vm/values/types.nim#L58

```text
    #  labels: vm, values, enhancement, open discussion

    # TODO(Vm/values/types) add new `:exception` type?
    #  this could work well with a new potential `try?`/`catch` syntax, 
    #  or a `throw` method
    #  labels: vm, values, enhancement, open discussion

    ValueKind* = enum
        Null            = 0
        Logical         = 1
ndex bc1d0e2e3e..522c56a310 100644
++ b/src/vm/values/value.nim
```

561a91b4b612d1fad1552bfd415aa8143db2de20

RickBarretto:
# Description

I was talking with @drkameleon about some design decisions in Arturo and then he came with a challenge to me: _write my first **type** in the project._

Well, challenge accepted, let's see what can I do! :wink:

- Fixes #1043, #1028

## Type of change

- [ ] Code cleanup


RickBarretto:
@drkameleon, I'll need your opinion about this initial draft.

drkameleon:
I've had a look into your initial draft.

I like several points, I have my doubts about other points (mainly the point that made me re-think the whole thing was the use of literals as Exception types; yes, it kind of makes sense, but I think I have an idea...)

So, I'll elaborate... :)

But, let's start by saying that I suggest the create of 2 new types: `:exception` & `:exceptionType` 🙄 .

Your use of literals for naming exception types *does* makes sense, but I'm not too sure it makes a lot of sense given how Arturo values work. I mean, we generally use the values *themselves*, not literals to refer to values (unless we are actually talking about in-place modification). Also, how can't we define some pre-set Exception's, pretty much like we define Quantities-related constants? We could have used e.g. `'planckConstant` and have the system look up the string to see what it means, but in fact `planckConstant` is itself a pre-defined constant; that is a Quantity value. (Still not sure I'm making any sense the way I put it, but ok... lol).

So, let's see what I could suggest

**The simplest, barebones version:**

```red
myFunc: function [x][
      if x = 2 -> throw "This is an error"
      return x + 3
]

if not x: <= try [myFunc 2][
      ; x is going to be an :exception
      print ["We got an error:" x] ; print a RuntimeError-styled error with the error message above
]
```

**The full-blown, more-complicated version**

```red
FileNotFound: to :exceptionType "File not found" 
; this type of things/constants could even be defined as such,
; meaning we could use them everywhere, plus system functions could
; even be made to enforced the exact same "exceptions"
; the string in there will work like the "identifier" of the exception
; and could even be shown e.g. as the first line of our error messages

readFile: function [filepath][
      if not exists? filepath ->
            throw.as:FileNotFound ~{Path: |filepath| doesn't appear valid}
      ; let's say we could set the "message" as the only must-have argument for throw
      ; if we don't set an `.as:` field it will be thrown as a GenericError - or sth like that - that is
      ; a default :exceptionType
            
      return read filepath
]

if not contents: <= try [readFile "somefile.txt"][
      ; x is going to be an :exception (an exception will contain two accessible fields - accessible via `get`/`/` I mean - that will be `\message` and `\type`

      if contents\type = FileNotFound ->
            print ["damn... We got this:" contents\message]
]

; I admit that... the more I'm looking at it, the more I like it... 
; it's quite Arturo-style, it keeps using our values system & syntax and 
; could actually mean we'll have Exceptions by type.... Hmm...
```

So, what I'm suggesting, trying to be as specific as I can be:

- Create a new Value type as: `ExceptionType` (or `ExceptionClass`? Actually, I wouldn't totally mind this one, since we don't use "class" for anything, while we *do* use "type"). This would simply have a `label` field as Nim string.
- Create a new Value type as: `Exception`, with 2 fields: `message` (a `string`) and `type`/`class` (a Value that will be pointing to an ExceptionType value actually)
- Add the necessary overloads for boring things like comparing the values (mainly, this would be valid only for `==`, since how could ExceptionA be > ExceptionB, so not a lot of work to be done here), etc - the way I usually do it is: add the types in [here](https://github.com/arturo-lang/arturo/blob/master/src/vm/values/types.nim) first, compile, and see what part of the code is complaining - fix that & repeat 😉 
- We'll obviously need a couple of Value constructors (for Exception & ExceptionType values) in here, e.g.: https://github.com/arturo-lang/arturo/blob/master/src/vm/values/value.nim#L400 - but you'll actually know what type of constructors will need a bit later on while implementing the new functions
- Add a new module inside or library: **Exceptions** (the plural serves in that there won't be any naming clash)
- Make sure you actually import the new module! -> https://github.com/arturo-lang/arturo/blob/master/src/vm/vm.nim#L81-L105
- And then: create a new `throw` function in there (taking e.g. a `message` string argument, and an optional attribute `.as:` which will be an ExceptionClass.
- Also, move the existing `try` function from Core to Exceptions
- `try?` could be eliminated too
- If you feel extra-adventurous, in order for our new Exception value's `message` & `class` fields to be available like `exc\message`, you'd also have to add support for it here: https://github.com/arturo-lang/arturo/blob/master/src/library/Collections.nim#L629 (I'm warning you, the water is getting... deeper here, so we have to totally make sure we don't break anything, but it's still an interesting challenge! 😉 )
- And, last but not least, you could add support for `to :exceptionClass <string>` conversions in here: https://github.com/arturo-lang/arturo/blob/master/src/library/Converters.nim#L85

The way I've written it all together, I think it sounds far more daunting than it is. But I actually think it's something you can perfectly tackle! 🚀 

And of course, I'm here for feedback and any sort of assistance - direct or indirect. 😉

RickBarretto:
@drkameleon, wow, thanks for this!
I particularly liked the idea of two separated types, instead of use a hack-ish `:literal` to represent this.

Well, about the `class`, I don't think this would be a good idea to don't make things confusing to new users. In fact, our exception type is not a class-based one as it's for Python or Ruby...

What we would use is `exceptionKind` and `kind` field instead, IMHO.
Other idea would be to use `:error` instead of `:exception`, that is a shorter and still a good one.

RickBarretto:
What do you think about this?

```red
fileNotFound: to :exceptionKind "File not found" 

f: $[][
    throw fileNotFound
]

if err: <= try -> f [
    ensure -> fileNotFound = err\kind
    ensure -> "File not found" = err\message
]
```
Basically, `throw` can accept `:exceptionKind` for default message errors.

RickBarretto:
@drkameleon, I updated the file, say what you think about this new implementation.

drkameleon:
At a first glance, I like both your ideas: I prefer using `kind` (short and not confusing) and I also don't dislike your idea to call them `Error` values ;)

Let me look into the update carefully and I'll let you know! 😉 

Awesome stuff!! 🚀 🚀 🚀

drkameleon:
> What do you think about this?
> 
> ```red
> fileNotFound: to :exceptionKind "File not found" 
> 
> f: $[][
>     throw fileNotFound
> ]
> 
> if err: <= try -> f [
>     ensure -> fileNotFound = err\kind
>     ensure -> "File not found" = err\message
> ]
> ```
> 
> Basically, `throw` can accept `:exceptionKind` for default message errors.

🤔 

Twist on the same idea:

it could accept both strings (which it would `throw` as a generic error, with the given message) and `:exceptionKind` (or `:errorKind`). How about that?

drkameleon:
All that being said, regarding naming and what you've suggested, I think I would go with `:error` & `:errorKind` 😉

RickBarretto:
> How about that?

Sounds good to me!

> I think I would go with :error & :errorKind 😉

Sure! :wink:

RickBarretto:
I'm trying to figure out why I'm getting this error...

RickBarretto:
This is incomplete and will break on CI, I'll have a better looking tomorrow, now it's too late to me.

But what should I do with JsonNodes? @drkameleon.

drkameleon:
I've had a look into your changes and I'd say you are on the right track! 😉 🚀 

**Re: helpers/jsonobject**

Since it tries to handle all possible Arturo values when converting to Json, it's good practice to put them all somewhere even though we might not yet know what to do with them (after all, you realize that the chances of someone *actually* wanting to convert an error into Json are extremely slim). What we could possible do, is convert them both into some sort of string (which would be the `$` you have already added 😉 )

Now, let' see what errors you're getting ...

drkameleon:
I've just had a look into the errors I see here in the CI builds.

The main one (and probably one of the last ones you'll have to deal with) has to do with the Converters library module. Namely: https://github.com/arturo-lang/arturo/blob/master/src/library/Converters.nim#L98

Just add cases for `Error` & `ErrorKind` with a simple discard - basically add them at the very bottom - and you're good to go! Btw, this routine is the one handling all `to :whatever something` conversions ;)

Great stuff! Keep going! :)

drkameleon:
Also, for reference (in case you haven't seen it I mean), you may have a look at this: https://github.com/arturo-lang/arturo/pull/1308

I added a new `PathLiteral` type yesterday and did my absolute based to isolate and document all commits so that you could - potentially - have a look and/or use it as a guide... Just saying! 😉

RickBarretto:
> (which would be the $ you have already added 😉 )

I did this, but I've got an error, so I took a step back.

> Just add cases for Error & ErrorKind with a simple discard 

Alright!

> Also, for reference (in case you haven't seen it I mean), you may have a look at this: #1308

I forgot to mention, Yes I'm using this as a reference for me. This is helping me a lot.

drkameleon:
I know the PR is still in progress, but I decided to take a look at it anyway...

Good job in general - I think it has started taking shape + it does look good! 😉 

One little thing I noticed (although a consistent one) and I'm mentioning it since it's actually an error (a silent one, but still an error):

E.g.
https://github.com/arturo-lang/arturo/blob/c84acaa8710678d59bdffb47305cf40cca4b6a4e/examples/rosetta/detect%20division%20by%20zero.art#L2-L3

That was before:

try? [
]

RickBarretto:
About the `throws?` function, it does exactly the same thing as `try?` (at least at user level), but the first one fits better than `try?` semantically.

drkameleon:
> About the `throws?` function, it does exactly the same thing as `try?` (at least at user level), but the first one fits better than `try?` semantically.

Totally! 😉

RickBarretto:
TODO:
- [ ] Implement `to :errorKind` from `:string`
- [ ] Implement the `throw` function
- [ ] Implement unit-tests

RickBarretto:
This will break, since `try` returns an `:error` now. I need to update unit-tests later.

RickBarretto:
@drkameleon , what do you think?
If you think it's good, I'll finish implementing more tests and the proper documentation usage.

examples/_drafts/parser.art
drkameleon, line 1: I had totally forgotten about this! Good to know it's around... lol

src/library/Collections.nim
drkameleon, line 1: When I saw code edits in Collections\get, I was like... "what is that for?!". But actually it makes perfect sense + the implementation is exactly along the lines of what I would have done myself! 😉

src/vm/values/custom/verror.nim
drkameleon, line 1: Love this 😉

src/library/Exceptions.nim
```diff
@@ -0,0 +1,148 @@
...
+import vm/values/custom/verror
+
+proc defineSymbols*() =
+
+    constant "genericError",
+        alias       = unaliased,
+        description = "A generic :errorKind":
+            newErrorKind(verror.genericErrorKind)
+
+    builtin "throw",
```
drkameleon, line 16: I can see some asterisks here, I mean regarding how our VError is raised (or to be more precise: how it's processed - later on, which it isn't), but I think I'll handle this myself 😉

drkameleon (changes requested):
I've had a look through the whole PR and I can say I generally love it.

There might be issues with how errors show up (and I will change a few things about it), but the idea is spot on. 

(Basically, I've been thinking that I could make *all* of our errors - the errors Arturo emits - comply with our `VError`, so I'll end up re-writing a good part of that code, that would help a) for reasons of consistency, b) auto-magically resolve the output issue).

Only one thing I'd ask is if you can add 2 predicates in the Reflection module? (one for `error?` and another one for `errorKind?` - since we tend to have one for each of them, let's keep it consistent) - it's very easy to implement (shouldn't take you more than 5-10 minutes), but it's good for you so that you can add the new types all the way! 😉 

Also, tiny sidenote: given the cleanup I have done (still in the PR, but I've been waiting for this one to be merged first), the predicates (e.g. `integer?`, etc) are all now in a Types module, along with `define` for example, and keeping it all organized in a (hopefully) much better way, but just don't bother about it - I'll deal with the conflicts mess myself! lol ;-)

----

All that being said, after you add the 2 little functions, I'm merging it - so that we can move on. And we can both work on optimizations later on based on the master branch! 🚀 🚀 

Great job! :)

src/library/Exceptions.nim
```diff
@@ -0,0 +1,148 @@
...
+{.used.}
+
+import vm/lib
+import vm/exec
+import vm/errors
+import vm/values/custom/verror
+
+proc defineSymbols*() =
+
+    constant "genericError",
```
drkameleon, line 11: Re: this.

Just noticed it too. I love it that you included an error as a constant (to get it started). Now, why do I keep thinking that I might like something like `GenericError` for this type of thing? (I'm not telling you to do - or not do anything - I'm just thinking out loud, kind of :) )

RickBarretto:
> (shouldn't take you more than 5-10 minutes),

Less than 2 minutes to be more precise.

> but just don't bother about it - I'll deal with the conflicts mess myself! lol ;-)

Alright, no prob! 😉

PR #1305, opened 2023-11-20
Make object functions callable

drkameleon:
# Description

Right now, if we have a function field within a Dictionary value, this can be perfectly called as a normal function (albeit in a different execution level, but still...):

```red
myDict: #[
     myFunc: function [x][
           print ["x:" x]
     ]
]
do ::
     myDict\myFunc 10 ; this works perfectly
```

If `myDict` was an Object (which - let's not forget - is still a dictionary internally, only presented as a different type and a few bells and whistles), this wouldn't work.

**For example:**

```red
define :myObj [x][
      init: [
            this\myFunc: function [x][
                   print ["x:" x]
            ]
      ]
]

MyObj: to :myObj [2]
do ::
      MyObj\myFunc 10
      ; this could easily work - not sure how far we can take it and if we could actually wrap `this`
      ; so as to make it approximate some more "real" OOP, but still... this should simply work
```

## Type of change


drkameleon:
Ok, so... initially we've done what this PR was for (in like 3 lines lol).

Now, let's see if we can take up the challenge of actually having callable object functions *with* `this` (this one is far more challenging than it sounds, but let's see...) For now, I'd say it would be better to leave that for another PR.

So...

Ready to merge! 🚀

PR #1308, opened 2023-11-22
Implement new PathLiteral type

drkameleon:
# Description

Right now, we have `:path` (e.g. `a\b\c`) & `:pathLabel` (e.g. `a\b\c:`) values, but we have no way to refer to them as literals.

**What I mean is quite simple:**

- if we have a function `f` and we want to e.g. have a look at its info, we'd do: `info 'f`
- if we have a function hidden in a dictionary, or an object method, e.g. `obj\method`, we should be able to do something like: `info 'obj\method` (it's consistent with our syntax as a whole + right now, there is no way at all AFAIK to access this type of information)

Also, @RickBarretto I'll try to annotate the commits carefully so that you may use this as a reference [in your own PR](https://github.com/arturo-lang/arturo/pull/1304), as to how we normally add a new value type (this one, will also include a new syntax for the type - which means it will involve additions to the parser - but, other than that, the process will be very similar 😉 )

## Type of change


drkameleon:
And... 

ready to merge! 🚀

PR #1314, opened 2023-11-25
Revamp Arturo's OOP capabilities

drkameleon:
# Description

As of now, [`define`](https://arturo-lang.io/documentation/library/converters/define/) already... defines an `.as` option that was supposed to be used for declaring the "parent" type (= the type our new type is meant to inherit). However, this was never really implemented.

That's exactly what this PR is about. Let's go! 🚀  (and hope that I don't break it :P)

**Also see:**

- https://github.com/arturo-lang/arturo/pull/1315
- https://github.com/arturo-lang/arturo/pull/1316
- https://github.com/arturo-lang/arturo/pull/1318

**Fixes:**

- [x] Fixes #1125
- [x] Fixes #1323 

(and many more)

## Type of change

- [x] Code cleanup
- [x] Enhancement (implementation update, or general performance enhancements)
- [x] New feature (non-breaking change which adds functionality)
- [x] Documentation (documentation-related additions)

PR #1315, opened 2023-11-27
Fine-tune object-oriented aspects of Arturo

drkameleon:
# Description

Given that Arturo's user-defined types, went from being simple Dictionary values (with a twist and 3 potential "magic" functions) to something more (namely: a scoped wrapper for object functions - methods - with access to `this` ❗,etc), I think it's high time we did a couple of things "properly".

- the already-existing `.as:` option (for inheritance) that has been there for some time has to be working properly - this has mainly been dealt with in the [parent PR ](https://github.com/arturo-lang/arturo/pull/1314)
- since we can now define type-specific functions that work with an injected `this`, why should `init`/`print`/`compare` remain something magic and random and not just be proper functions as well
- `init` right now works *after* the object has been initialized, so it's not a proper constructor
- also, the magic fields that are being declared as part of [`define`](https://arturo-lang.io/documentation/library/converters/define/)'s 2nd parameter make the whole thing quite confusing. we could keep something like that only as an option (e.g. `.with:`) where the fields being declared we create an automated `init` method that will simply take the values and assign them to the newly created object.
- all that being said, the new `init` will have to work exactly like all other function (including the "magic" ones: `print`/`compare`) in the sense that: if there is a new implementation declared inside a type definition, that's the one to be used (= override parent method); otherwise we'll be using the one of its parent - if any
- also, this last point implies that we must somehow have access to a type's parent (or at least the type's parent constructor)

So, a few scattered thoughts that - I hope will make a difference.

Again: the goal is not to suddenly make Arturo object-oriented. Just that, since we do have the option of defining custom types/classes, why not do it a tiny bit more... correctly? 😉 

## Type of change

- [x] Code cleanup


drkameleon:
To avoid make any messier, with chains and chains of PR one on top of another, let's finalize this too... and we'll fix pending issues starting with its parent PR.

That being said...

Ready to merge (too)! 🚀

PR #1318, opened 2023-11-30
Add support for automated type constructors

drkameleon:
# Description

**The current (master branch) version is:**

```red
define :person [name, surname][]
```

And then - after `a: to :person ["John", "Doe"]` - you may access (read/write) both fields, like  e.g. `a\name` or `a\surname`.

**The current, non-merged (yet), parent-PR equivalent would be:**

```red
define :person [
     init: function [name, surname][
          this\name: name
          this\surname: surname
     ]
]
```

Obviously, from being too barebones (totally non-OOP), now it seems as if we've suddenly gone OOP all the way. So, let's do something about it: what if the passed block to the definition is not a dictionary (or a dictionary-able block) but a block with Word values, etc (as before)? Then, we'll automatically generate the above, but following more or less the syntax that is already there (minus the 2nd parameter which won't be there...).

**What we'll do here:**

```red
define :person [name, surname]
```

And... boom! 💣 The code is equivalent to the 2 versions above!  
(Obviously, that doesn't work - yet. That's what this PR is for! 😉 )

## Type of change

- [x] Code cleanup


drkameleon:
Despite the pending issues and the CI builds failing (mostly unit-tests & still-not-updated RC examples), I think we've covered what we were supposed to do here; and so more!

So...

Ready to merge! 🚀

PR #1344, opened 2023-12-19
Add new `:method` type

drkameleon:
# Description

**What are `method`s supposed to be?**

Basically, `function`s that are within a new type. (defined via `define`)

**Can we not do it with old-school `function`s?**

We sure can. The problem is:

- There are function options that are totally irrelevant in the case of object methods (e.g. `.import`, `.export`? not really the point...) and there is also too much unnecessary baggage (e.g. there is no way a method to be in the stdlib)
- They have to be able to support `this`-injection, as-automatically-as-possible
- It might allow to correctly inject this at runtime, during AST processing - something which is not possible right now, especially for nested calls (method within a dictionary)
- It will allow us to have a constructor (e.g. `method`) that in contrast to `function` (where it would be totally irrelevant) will allow for more specific options (e.g. `.new` - or `.override`? - to allow for explicitly specifying magic methods)

## Type of change


drkameleon:
This isn't 100% ready yet either... but let's merge back to the parent to keep track of things, until we resolve the few pending issues.

So...

Ready to merge! 🚀

PR #1350, opened 2023-12-21
Add more "magic" type methods

drkameleon:
# Description

Right now, we have a couple of magic methods that can be defined within a new type that affect how things are done in different ways. E.g. defining a `compare` for a specific type, makes `<` work for objects of that type, or defining `print` (I'd change that as a name!) makes stringfication (`print` or `to :string`) work.

So, can we add some more? Let's see! 🚀 

## Type of change


drkameleon:
Let's merge backwards and continue from there! 😉 

So... ready to merge! 🚀

PR #1353, opened 2023-12-23
[VM\value\comparison]: Clear comparisons with `getValuePair`

### LINKED ISSUE #1149: [VM/value/comparison] rewrite all `==`, `<`, `>` methods to use `getValuePair`?
[VM/value/comparison] rewrite all `==`, `<`, `>` methods to use `getValuePair`?
This would follow closely the implementations found in VM/values/operators
This could be a good idea for cleaner code, but we should thoroughly benchmark it before!

https://github.com/arturo-lang/arturo/blob/269a723e23971f7c5ba310f6c3539852a6f016e3/src/vm/values/comparison.nim#L56

```text

# TODO(VM/values/comparison) Should we throw errors in case of incompatible pairs?
#  For example, if we try to compare a string with a number, should we throw an error?
#  Or should we simply return false?
#  Also, this should affect `<` and `>`, more than `==` since - at least IMHO - something
#  like `2 = "hello"` still *does* make sense (it may be obviously false, but it does). 
#  `2 < "hello"`, for example, does *not*.
#  see also: https://github.com/arturo-lang/arturo/pull/1139
#  labels: enhancement,values,error handling,open discussion

# TODO(VM/value/comparison) rewrite all `==`, `<`, `>` methods to use `getValuePair`?
#  This would follow closely the implementations found in VM/values/operators
#  This could be a good idea for cleaner code, but we should thoroughly benchmark it before!
#  labels: enhancement, cleanup, benchmark,performance, values

# TODO(VM/values/comparison) Verify all value types are properly handled by all overloads
#  labels: vm, values, enhancement, unit-test

```

d9a0f47742c8e76413ea645c1a62672e8b0cf704

RickBarretto:
# Description

Benchmarks and tries to replace the ugly `case`/`of` by our newer, faster? and more readable approach `getValuePair`.

- Fixes #1149

## Type of change

- [x] Code cleanup

### DIFF
--- a/bench
+++ b/bench
@@ -2,7 +2,15 @@
 
 git log -n 1 --pretty=medium > bench-log
 
-{ time ./bin/arturo.exe ./tools/tester.art; } 2>> bench-log
+echo -e "\nOLD" >> bench-log
+{ 
+    time ./bin/reference.exe ./tests/unittests/lib.comparison.art; 
+} 2>> bench-log
+
+echo -e "\nCURRENT" >> bench-log
+{ 
+    time ./bin/arturo.exe ./tests/unittests/lib.comparison.art; 
+} 2>> bench-log
 
 git add bench-log
 git commit -m "[bench]: update benchmark's log"
--- a/bench
+++ b/bench
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+git log -n 1 --pretty=medium > bench-log
+
+{ time ./bin/arturo.exe ./tools/tester.art; } 2>> bench-log
+
+git add bench-log
+git commit -m "[bench]: update benchmark's log"
--- a/bench
+++ b/bench
@@ -5,14 +5,14 @@ git log -n 1 --pretty=medium > bench-log
 
 echo -e "\nOLD" >> bench-log
 { 
-    time for counter in {1..50} 
+    time for counter in {1..250} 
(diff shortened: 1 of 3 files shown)


RickBarretto:
# Results

* Running 250x for each version

```
commit 1ec6347477e5e646aae23a1e677862705c5f447b
Author: RickBarretto <78623871+RickBarretto@users.noreply.github.com>
Date:   Sun Dec 24 18:51:12 2023 -0300

    [VM\values\comparison]: minor fix

OLD

real	9m2.314s
user	0m0.332s
sys	0m0.809s

CURRENT

real	8m35.180s
user	0m0.181s
sys	0m0.842s
```

RickBarretto:
@drkameleon, feel free to test and benchmarking it.
If you think it's good to merge, please, don't forget to delete the files `bench` and `bench-log`.

Thanks, and Merry Christmas! 🎄

drkameleon (approved):
Wow! That's some extremely good work here! 🚀 

Very nicely done. That was exactly the spirit when I came up with `getValuePair` - I used it for operations, but forgot (or better put: I waited until I made sure it wasn't too bad) to use it for other cases, like comparison.

Pure awesomeness! 😄

drkameleon:
Ready to merge! 🚀

PR #1432, opened 2023-12-28
Sanitize error system

drkameleon:
# Description

Right now, the errors produced are quite messy, or incomplete, or not consistently-presented.

At the same time, we have new Error & ErrorKind values.

The two system should - ideally - be merged into one. Or at least that's what we'll attempt here.

Let's go! 🚀 

- [x] Fixes: #1561 
- [x] Fixes: #1562 

----

## Type of change

- [x] Code cleanup


drkameleon:
The PR is not 100% complete - but let's say... it's "complete enough".

Many of the original goals have been completed and the whole error system is practically brand new. Now, do I like every single error message? Nope, and that's why I'll follow with different PRs that are going to address the pending issues. 

Also, there are different errors that are (still) not getting caught or properly reported, etc. But... all in due time! 😉 

----

That having been said...

Ready to merge! 🚀

PR #1438, opened 2023-12-28
[Core\var]: add support for `:word`

### LINKED ISSUE #1325: Core\var: should also support Word values
Pretty much like #1324 😉

RickBarretto:
# Description

A short demonstration:
image

- Fixes #1325

## Type of change

### DIFF
--- a/src/library/Core.nim
+++ b/src/library/Core.nim
@@ -1105,7 +1105,7 @@ proc defineLibrary*() =
         rule        = PrefixPrecedence,
         description = "get symbol value by given name",
         args        = {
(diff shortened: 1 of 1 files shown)


drkameleon (approved):
Simple but nicely done!

Basically, the addition looks weird-ish, but the use case scenario is basically you're last example, e.g. `loop [a b] ...`. It's probably not the most common use case, but when you simply have that, why not be able to simply use the Word value (which is *already* a string)?! So, there we go!

Great! 😉

drkameleon:
Ready to merge! 🚀

PR #1439, opened 2023-12-29
[vm\globals] ultra-experimental refactoring features... Enhancing control-flow 

RickBarretto:
# Description

I made this PR just as a ultra-experimental and drafty way of try new features for refactoring.

In this case, I removed some control-flow by using polimorphism with `FunctionFlag`. Something imitating [the Arturo's `:attribute`s][attrs] for procedures. Thanks @beef331 for your help with that.
And also, removed the code repetition when declaring variables inside an `if` statement, by copying [the Walrus Operator `:=` from Python (Formally, "Assignment Expression")][walrus].

[attrs]: https://arturo-lang.io/documentation/language/#attributes
[walrus]: https://peps.python.org/pep-0572/

## Type of change

- [x] Code cleanup

### DIFF
--- a/src/vm/globals.nim
+++ b/src/vm/globals.nim
@@ -122,11 +122,11 @@ proc FetchSym*(s: string, unsafe: static bool = false): Value {.inline.} =
     ## Checks if a symbol name exists in the symbol table
     ## - if it doesn't, raise a SymbolNotFound error
     ## - otherwise, return its value
-    when not unsafe:
+    when unsafe:
+        Syms[s]
+    else:
         if (result = Syms.getOrDefault(s, nil); unlikely(result.isNil)):
             RuntimeError_SymbolNotFound(s, suggestAlternative(s))
-    else:
-        Syms[s]
 
 proc FetchPathSym*(pl: ValueArray): Value =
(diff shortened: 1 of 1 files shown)


RickBarretto:
I just compiled locally, let's see if this runs...

RickBarretto:
@drkameleon, I'll want your views about that.
This is not an important PR, and I think these sugar features should be placed into another module...

drkameleon:
> @drkameleon, I'll want your views about that. This is not an important PR, and I think these sugar features should be placed into another module...

I'm open to anything as I've said that either enhances the performance or makes our code look cleaner.

That being said, I would be *extremely* carefully with what Nim does in the background. I mean... assignments are very, very risky and we should be 100% sure what it does, and if the assigned thing is e.g. a pointer, or not, since this might lead either to redundant copies or to things being changes in an unpredictable way. Also, since `globals` contains some of the most crucial functions, obviously everything should be double/triple-tested.

But I don't dislike the idea at all! 😉 

Also, moving them to a separate module makes sense to me to be honest. Perhaps, another `sugar` helper (in helpers/)?

PR #1440, opened 2023-12-29
[Collections\first]: stuck if given a .n: value higher than the size of the collection

### LINKED ISSUE #1335: Collections\first: stuck if given a `.n:` value higher than the size of the collection
Basically, this should return the whole collection instead of crashing silently.

**Example:**

```
first.n:6 @1..4
```

should return: `[1,2,3,4]`

RickBarretto:
# Description

Stuck if given a .n: value higher than the size of the collection

- Fixes #1335

## Type of change

### DIFF
--- a/src/library/Collections.nim
+++ b/src/library/Collections.nim
@@ -615,7 +615,11 @@ proc defineLibrary*() =
             if checkAttr("n"):
                 if xKind == String:
                     if x.s.len == 0: push(newString(""))
-                    else: push(newString(x.s[0..aN.i-1]))
+                    else:
+                        if aN.i > x.s.len:
+                            push(newString(x.s[0..^1]))
+                        else: 
+                            push(newString(x.s[0..aN.i-1]))
                 elif xKind == Range:
                     if aN.i == 1 or aN.i == 0:
                         push(x.rng[1, true])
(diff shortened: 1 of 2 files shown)


src/library/Collections.nim
```diff
@@ -615,15 +615,26 @@ proc defineLibrary*() =
             if checkAttr("n"):
                 if xKind == String:
                     if x.s.len == 0: push(newString(""))
-                    else: push(newString(x.s[0..aN.i-1]))
+                    else:
+                        if aN.i > x.s.len:
+                            push(newString(x.s[0..^1]))
+                        else: 
+                            push(newString(x.s[0..aN.i-1]))
```
drkameleon, line 622: I'm seeing some CI build fails, so I'm not sure if I'm supposed to... "review" this one, but I go ahead and do it either way lol 🤣 

The code looks good, although I'm wondering if we could use e.g. [`min`](https://nim-lang.org/docs/system.html#min%2Cint%2Cint), instead of these different `if`-`else`s (performance-wise it should make a tiny bit of a difference, since `min` is a Nim "magic" proc, plus it could look cleaner... just saying 😉 )

Like: 
```suggestion
                        push(newString(x.s[0..min(aN.i,x.s.len)-1]))
```

What do you think?

drkameleon (changes requested):

PR #1447, opened 2023-12-31
[Iterators]: check block value for  `fetchParamsBlock`

### LINKED ISSUE #1137: [Iterators] we should make sure that params block is either literals or words
[Iterators] we should make sure that params block is either literals or words

https://github.com/arturo-lang/arturo/blob/5fb5c1d23d0fcc73685ae17ac47f632b4a049fea/src/library/Iterators.nim#L272

```text
    finalizeLeakless()

# TODO(Iterators) we should make sure that params block is either literals or words
#  labels: library, enhancement, error handling, performance
template fetchParamsBlock() {.dirty.} =
    var params: seq[string]
    if hasIndex: params.add(withIndex.s)
ndex bd7c69185f..91d14b776d 100644
++ b/src/library/Numbers.nim
```

26c1a156e4e7654b01f78277ff4a420d7045d4d2

RickBarretto:
# Description

Makes sure that parameters block is either literals or words

- Fixes #1137

## Type of change

- [ ] Code cleanup

### DIFF
--- a/src/library/Iterators.nim
+++ b/src/library/Iterators.nim
@@ -277,6 +277,7 @@ template fetchParamsBlock() {.dirty.} =
     if hasIndex: params.add(withIndex.s)
     if yKind != Null:
         for item in mitems(y.a):
+            item.requireValue({Literal, Word})
             params.add(item.s)
 
 template prepareIteration(doesAcceptLiterals=true) {.dirty.} =


RickBarretto:
I've tried with `requireValue` instead, let's see!

drkameleon (approved):
Your idea is spot on! Basically, that's exactly how we should be thinking all the time and not get carried away (as I did).

Now, this is just a test *before* the iteration, so it shouldn't add toooo much of a delay, but still... (if we are talking about iterations within iterations, then it would!).

So, any performance gain is more than welcome - and we don't have to iterate over the same block twice.

100% well done! 🚀

drkameleon:
Ready to merge! 🚀

PR #1516, opened 2024-01-10
Safe in-place overflows in arithmetic operations

drkameleon:
# Description

The maximum "native" number supported by Arturo is a 64-bit `int` (or to be precise: 9223372036854775807). If we move outside this range, the number is to be automatically converted to a GMP-powered BigNum. And all this happens silently, without the user having to bother whether the underlying thing is a 64-bit int, or a GMP `mpz_t`.

To do this, we have to be able to track operation overflows correctly. And - generally - taking for granted that an operation between two int64's will be valid is _totally wrong_ ⚠️ .

**Quick example:** 9223372036854775807 + 1.

The operands are both int64, but the result will actually overflow our type.

To predict it, we have this absolutely wonderful set of intrinsics (and more): https://github.com/arturo-lang/arturo/blob/master/src/helpers/intrinsics.nim

And they are consistently used when appropriate, e.g. in our `+` implementation: https://github.com/arturo-lang/arturo/blob/master/src/vm/values/operators.nim#L169-L176

The problem we're using it exactly like that for in-place operations too: https://github.com/arturo-lang/arturo/blob/master/src/vm/values/operators.nim#L181-L185

And the problem lies in the fact that `addIntWithOverflow` operates with a pointer as a result. So, if it fails, it will have *already* changed our initial value; in a few words: whatever we do with it later on will be gibberish.

**An example of that, in practice:**

Screenshot 2024-01-10 at 17 15 30

And this actually affects both normal operations, and operations between Rationals (and as a result, Quantities too!).

So, let's fix that at last! 🚀 

Fixes # (issue/s)

## Type of change


drkameleon:
Ready to merge! 🚀 

---

P.S. Let's deal with pending issues in the parent PR, as usual... 😉

PR #1524, opened 2024-01-13
Fix web-builds

RickBarretto:
# Description

Basically, I was fooled by CI. This is building on Mini-mode instead of Web-mode. And needs fix ASAP.

- Fixes https://github.com/arturo-lang/nightly/issues/5

## Type of change

- [ ] Code cleanup

### DIFF
--- a/.github/workflows/web.yml
+++ b/.github/workflows/web.yml
@@ -70,7 +70,7 @@ jobs:
       - name: Build Arturo
         run: |
           echo "$HOME/.arturo/bin" >> $GITHUB_PATH
-          ./build.nims build --mode ${{ matrix.version }} --log
+          ./build.nims build --mode ${{ matrix.platform }} --log
 
       - name: Upload Artifacts
         uses: 'actions/upload-artifact@v2'
--- a/build.nims
+++ b/build.nims
@@ -465,9 +465,9 @@ cmd build, "[default] Build arturo and optionally install the executable":
             safeBuildConfig()
             miniBuild()
         >> ["web"]:
-            config.binary     = config.binary
-                                      .replace(".exe", ".js")
-            config.version    = "@web"
+            config.binary  = config.binary.replace(".exe", ".js")
+            config.version = "@web"
+            webBuildConfig()
             miniBuild()
 
     match args.getOptionValue("os", default=hostOS, into=availableOSes):
--- a/oldbuild.nims
+++ b/oldbuild.nims
@@ -1,0 +1,343 @@
+# initial conversion to NimScript thanks to:
+# - Patrick (skydive241@gmx.de)
+
+# TODO(build.nims) General cleanup needed
+#  labels: installer, enhancement, cleanup
+
+#=======================================
+# Libraries
+#=======================================
+
+import os, parseopt, sequtils
+import strformat, strutils, tables
+
+import src/helpers/terminal
+
+#=======================================
+# Initialize globals
+#=======================================
+
+mode = ScriptMode.Silent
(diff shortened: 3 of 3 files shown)


RickBarretto:
```
C:\Users\rickb\Downloads\nim-2.0.0_x64\nim-2.0.0\lib\pure\rationals.nim(103, 13) Error: illegal conversion from 'float' to 'int'
      ai = int(x)
              ^
??? compiler msg initiated here [MsgOrigin]
No stack traceback available
To create a stacktrace, rerun compilation with './koch temp js <file>', see https://nim-lang.github.io/Nim/intern.html#debugging-the-compiler for details
stack trace: (most recent call last)
```

The strange part, is that this is an error that I had a lot of times, months ago.
@drkameleon, do you know what this means?

RickBarretto:
I added the older builder file just to compare. I cleaned up some commands that are not used for test this.
I seems that I did not changes nothing about behavior.

drkameleon:
> ```
> C:\Users\rickb\Downloads\nim-2.0.0_x64\nim-2.0.0\lib\pure\rationals.nim(103, 13) Error: illegal conversion from 'float' to 'int'
>       ai = int(x)
>               ^
> ??? compiler msg initiated here [MsgOrigin]
> No stack traceback available
> To create a stacktrace, rerun compilation with './koch temp js <file>', see https://nim-lang.github.io/Nim/intern.html#debugging-the-compiler for details
> stack trace: (most recent call last)
> ```
> 
> The strange part, is that this is an error that I had a lot of times, months ago. @drkameleon, do you know what this means?

Tricky!

As I just wrote in the other message, I hadn't really tried building for Web (after the latest PR), so my guess is I might have to patch things up (esp. re: Rational/Quantity values) before it's ready to compile.

drkameleon:
I've been studying the issue for the past 2 hours... And the conclusion is not good :S.

Basically, I don't know how long it is that the web builds are not working, but apparently in the meantime one million different things are crashing.

I've made several changes to try disabling e.g. Quantity/Unit values (from web builds), but I keep coming across different issues. The Quantity-related issues with web builds seem to be addressable (not easily, but they could be). The problem is I come across sth new every second line... And honestly, I don't know.

We could try ironing everything out, but this could far more time than expected + not sure if it's worth the fuss at all (although "running in your browser doesn't sound that bad). Perhaps, add the changes in a different PR and have it as reference for the future, while we disable web builds altogether? 🤔 

Not really sure what I'd suggest with this one... but the amount of issues I'm seeing doesn't even seem to have to do with my own code - at times, I'm seeing issues or warnings that looks like bugs (on the Nim side of things).

Thoughts?

RickBarretto:
First, sorry for my late reply. For some reason, I'm not getting notifications of comments... I'll see how to config this ASAP.

In fact, web builds was always a shady thing, since we never actually unit-tested it. The tests were done manually and we expected it to have the same/similar behavior as `@mini` builds.

Well... I'll have a look tomorrow on it. I'll try to disable Quantities and Rationals.
Even this being something a bit shady to me (the js backend), I'll try to manage it.

If nothing can be done, I think a transpiler would be a good alternative. I could try do it, if you want to.
Some bonus are: we don't need to take care about pointers, bigints and bindings directly for the JS backend.

RickBarretto:
> Not really sure what I'd suggest with this one... but the amount of issues I'm seeing doesn't even seem to have to do with my own code - at times, I'm seeing issues or warnings that looks like bugs (on the Nim side of things).

😥

RickBarretto:
Closing this PR.
- Moved to #1527.

PR #1525, opened 2024-01-15
Attempt to fix WEB builds

drkameleon:
# Description

Related: https://github.com/arturo-lang/arturo/pull/1524

## Type of change


RickBarretto:
@drkameleon, any success trying to manage this?

drkameleon:
> @drkameleon, any success trying to manage this?

With absolute honesty, I haven't done anything with this so far; but I'm planning to give it another go.

At best, at least for now, the way I see it is we may be left without Quantity/Unit value in the Web version. But let's see if we can get the rest... I'll post the updates ;)

PR #1527, opened 2024-01-17
Fix web builds and cleanup

drkameleon:
# Description

...

## Type of change

- [x] Code cleanup


drkameleon:
@RickBarretto If you have some time and can confirm this is - more or less - ok (at least compiling for you too & don't see anything weird related to the build script/workflow file, for example), I would be grateful! 😉

RickBarretto:
image
Compiled! 🙂

RickBarretto:
@drkameleon, bad news:
IDK if I did something wrong, but some functions are not running:
image

drkameleon:
> @drkameleon, bad news: IDK if I did something wrong, but some functions are not running: image

Well, it's not as bad as it looks.

I mean... it *is* bad, but let me explain:

First of all, the Web version has always been meant to be a "light" version. And it already had its fair share of issues (I mean before the ones we've introduced this year, the previous release had its known - and not so known - issues).

That being said, for me, it's still a success. Are there issues? There are. But I'd say we'd better track them down (and solve them) on a case-by-case basis. (⚠️ These ones are going to be a lot, but we'd have to start from somewhere! 😉 ).

So, I'd say... we merge this one and experiment with how we could unit-test it (via CI too). Once the unit-tests start working, we may start dealing with the issues that we've found.

What do you think?

RickBarretto:
> What do you think?

😉, Alright!

drkameleon:
Ready to merge! 🚀 

----

It would be great if we opened issues for any Web-related problem we find (and let's hope there aren't - too - many lol)

PR #1529, opened 2024-01-18
Minimize web builds properly

drkameleon:
# Description

Currently, we have `uglify-js` for some basic minification (although I'm not sure if it's properly used) + a custom post-processing stage (mostly with string replacements).

Let's see if we can reinstate everything as it was and make it a tiny bit smoother. Even for non-browser use, 8.4MBs of JavaScript code doesn't sound very... acceptable - to put it mildly. lol

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #1569, opened 2024-02-12
Remove `unidecode` support

### LINKED ISSUE #1178: Completely eliminate unicode transcribing capabilities?
I've been looking into Arturo's binary size (and what the different components "weigh").

Apparently, there is one concrete feature that could be accounting for close to 25% of the binary size that is... barely used (on my Mac, out of ~9MBs, 2MB is due to the **[`unidecode`](https://nim-lang.org/docs/unidecode.html)**-related features)

**So, where do we currently use it?**

- In [`convertToAscii`](https://github.com/arturo-lang/arturo/blob/master/src/helpers/strings.nim#L26) in the Strings helper, which is in turn used in the Converters module (only [for `as.ascii`](https://github.com/arturo-lang/arturo/blob/master/src/library/Converters.nim#L740))
- In the [`unisort` helper](https://github.com/arturo-lang/arturo/blob/master/src/helpers/unisort.nim#L120), or basically when we try to `sort.ascii` (which *I* am not even sure what it actually does lol)

In a few words: we barely use it and even I have personally *never* used it.

**How do we solve this?**

One way - and the "easiest" one by far would be to simply eliminate it.

Another idea would be to make this into a module/package, once Arturo's package system is ready. (as a matter of fact, it could be a very nice idea of a first experiment/package)

Another idea would be to make it so that the actual data is not loaded until we need it (e.g. via [`loadUnidecodeTable`](https://nim-lang.org/docs/unidecode.html#loadUnidecodeTable%2Cstring)). The problem is *where* is this data supposed to come from? Will this be a dependency? A file in Arturo's installation folder?

In any case, wasting 2 out of 9 MB for that doesn't make any sense to me. (Being batteries-included is one thing and I'm all for that, but that's pretty much like packing Brittanica... just-in-case somebody needs it lol).

drkameleon:
# Description

See also: https://unidecode.pkgr.art/

Fixes #1178 

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #1600, opened 2024-04-10
Minor Package Manager-related additions

drkameleon:
# Description

Add support for an optional `executable` field in our package `info.art` files:

If the field is set (and points to a file within our package repo, for example) then a `<package-name>` "executable" will be created within `$HOME/.arturo/packages/bin` pointing to that file (= basically it will execute it using Arturo, via Bash), so... we could do `dummy` and get our pseudo-binary to run from anywhere. (provided the $PATH variable is correctly set, obviously).

For now, unfortunately this will work on Linux/macOS/BSD only, but I'm 100% open to see how we could make that work on Windows too! 😉 🚀 

## Type of change


drkameleon:
Ready to merge! 🚀

PR #1609, opened 2024-04-23
FIX: Error reporter crashing when not in REPL mode

drkameleon:
# Description

When an error message is to be shown, there seems to be an issue with the `printCodePreview` method not correctly loading the appropriate Frame (and thus, the corresponding path).

## Type of change


drkameleon:
Ready to merge! 🚀

PR #1613, opened 2024-04-26
Introduce `.public` methods

drkameleon:
# Description

The goal of this PR is to enable us to annotate a given method as `.public` (to be read as: "exportable"/accessible from the outside).

Not sure if/when this will be enforced on a user-type level (types `define`d by the user) - at least, yet - but it could most definitely be used in a future `module` function implementation. 🚀 

## Type of change


drkameleon:
Ready to merge! 🚀

PR #1620, opened 2024-04-29
Core\with: Add support for Dictionary values

drkameleon:
# Description

[**`with`**](https://arturo-lang.io/documentation/library/converters/with/) right now supports "injecting" a given symbol within a block (very practical for closure-related manipulations, etc). But what if we want to "inject" a value with a different name?

Then, we could use a dictionary! 😉 

## Type of change


drkameleon:
Ready to merge! 🚀

PR #1625, opened 2024-05-03
Fix AST handling of method calls

drkameleon:
# Description

Right now, this:

```ruby
define :person [
    init: constructor [
        name :string
        surname :string
    ]

    check: method [][
        inspect this
    ]
]

define :personWrapper [
    init: method [n, s][
        \pers: to :person @[n, s]
    ]

    doSth: method [][
        \pers\check
    ]
]

a: to :personWrapper ["John" "Doe"]!
a\doSth

b: to :personWrapper ["Jane" "Doe"]!
b\doSth
```

leads to this:

```
[ :person
        name     :        John :string
        surname  :        Doe :string
]
[ :person
        name     :        John :string
        surname  :        Doe :string
]
```

The cause is that `this\pers` inside `doSth` has already been embedded the first time the function was evaluated. And since - for performance reasons - we tend to store evaluated blocks (apparently, a bit too much), the next time this function is called, with a different object, the same one is getting served 🤦🏻‍♂️ .

So, let's fix it! 

## Type of change

### DIFF
--- a/src/vm/ast.nim
+++ b/src/vm/ast.nim
@@ -967,7 +967,7 @@ proc generateAst*(parsed: Value, asDictionary=false, asFunction=false, reuseArit
     var res = newRootNode()
 
     PipeParent = nil
-    CanStore = false
+    CanStore = true
 
     when not reuseArities:
         TmpArities = collect:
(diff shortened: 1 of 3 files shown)


drkameleon:
Ready to merge! 🚀

PR #1646, opened 2024-05-25
Start choose migration

RickBarretto:
# Description

This PR is not meant to be an implementation of the `choose` function itself, and this does not remove the target functions that will be deprecated soon.

Basically, the way `if?`, `else`, `case`, `when?` works is very problematic. Since they depends on each other, and should be independent. Other problem is the fact that `else` and `when?` violates the way the stack really works, so this is a valid code:

```art
false
else -> print "Whut??"
; Whut??
```

And as we may know, this can lead to hidden bugs in a lot of code. That is the reason why me and @drkameleon were avoiding for a long time to use those functions.

So, this is time to change. But first things, first.
We need tests first, to check if the behavior is what we desire, then we need to implement that function and then we need to optimize it.

Worth to say, that `choose` will replace all of them into once.

## Type of change

- [ ] Code cleanup

### DIFF
--- a/examples/_drafts/euler14.art
+++ b/examples/_drafts/euler14.art
@@ -7,11 +7,10 @@ loop 2..1000000 'n [
 	x: n
 	count: 0
 	while [and? x <> 1 x >= n][
-		if? even? x [
+		(even? x)? [
 			x: x / 2
 			count: count + 1
-		]
-		else [
+		][
 			x: (1 + 3*x) / 2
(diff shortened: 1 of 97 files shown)


RickBarretto:
I changed a lot of examples at once, I hope I didn't break anything... :gezz:

RickBarretto:
Ok, cancelling this PR...
image

PR #1650, opened 2024-06-13
Add better support for SemVer spec

drkameleon:
# Description

Right now, Arturo's Version values *do* support SemVer (v2.0).

However, we could add some better support for things like prerelease tags, etc.

In a sense:

- `1.2.3-alpha` should come *before*
- `1.2.3`

as per [the official specification](https://semver.org/).

And, for that, we'd better make sure we can tell apart the `-prerelease` part from anything following `+build.number` (and which shouldn't be taken into account for precedence reasons).

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #1690, opened 2024-09-17
[examples\100 door] refactoring problem

RickBarretto:
# Description

Just a minor refactoring... 🙂 

## Type of change

- [ ] Code cleanup

### DIFF
--- a/examples/rosetta/100 doors.art
+++ b/examples/rosetta/100 doors.art
@@ -1,11 +1,8 @@
-isOpen: map 1..101 => false
- 
-loop 1..100 'pass ->
-	loop (range.step:pass pass 100) 'door [
-		isOpen\[door]: not? isOpen\[door]
-	]
+; By: RickBarretto
+; Based on the old version by: DrKameleon
 
-loop 1..100 'x ->
-	if isOpen\[x] [
-		print ["Door" x "is open."]
-	]
+perfectSqrt?: $[x] -> equal? 0 (mod (sqrt x) 1) 
+
+loop 1..100 'door -> 
+	if perfectSqrt? door -> 
+		print ~"Door |door| is open."


RickBarretto:
Theoretically, this is fine...

```
- Running: 100 doors                                      <-------------------
Door 1 is open.
Door 4 is open.
Door 9 is open.
Door 16 is open.
Door 25 is open.
Door 36 is open.
Door 49 is open.
Door [64](https://github.com/arturo-lang/arturo/actions/runs/10894507944/job/30231339960?pr=1690#step:3:65) is open.
Door 81 is open.
Door 100 is open.
------------------->
[  ERROR  ]
```

RickBarretto:
I see this later...

drkameleon:
Funny enough, I think with your PR you have discovered an extreme case of hidden "bug" (?) kind-of:

the reason because the CI builds are failing is that the `.res` file contains extra spaces at the end of each line. And, yes, we are stripping spaces but not per-line; this happens for the leading/trailing spaces.

So, the question would be: why does the `.res` file have those spaces?

Let's look at the previous and current implementation: yours uses `print` with a string and mine used `print` with a block. Hmm, hmm, hmm... You see what I'm getting at? 🤣 

I'll investigate a bit, but obviously this shouldn't be like this!

RickBarretto:
@drkameleon TBH I already knew this error, but since I made this PR from the github dev, I didn't updated the res file.

Every time I create a new test that needs res file I need to override it. Sometimes not due to whitespace, probably some invisible character or encoding, IDK...

But, if you do it on Windows, you'll get a new problem because of the carriage return of the Windows. To fix that, I need to pipe the output from the bash instead of powershell...

But, oh, well...
That is one reason why I created unitt.

RickBarretto:
@drkameleon TBH I already knew this error, but since I made this PR from the github dev, I didn't updated the res file.

Every time I change some test that needs res file I need to override it. Sometimes not due to whitespace, probably some invisible character or encoding, IDK...

But, if you do it on Windows, you'll get a new problem because of the carriage return of the Windows. To fix that, I need to pipe the output from the bash instead of powershell...

But, oh, well...
That is one reason why I created unitt.

drkameleon:
> @drkameleon TBH I already knew this error, but since I made this PR from the github dev, I didn't updated the res file.
> 
> Every time I change some test that needs res file I need to override it. Sometimes not due to whitespace, probably some invisible character or encoding, IDK...
> 
> But, if you do it on Windows, you'll get a new problem because of the carriage return of the Windows. To fix that, I need to pipe the output from the bash instead of powershell...
> 
> But, oh, well... That is one reason why I created unitt.

Hmm... Interesting.

The problem with hidden characters is one thing. I mean it may happen for various reasons (including the editor, where you copy/paste from, etc).

Now, the thing with the spaces, I'll have to test it for myself.

Last but not least... regarding carriage return... I don't know. Perhaps, we should find a way to normalize this across the different OS'es?

Meaning:

```
something
something else in a different line
```

I think that comparing this - as a string - to itself, in different OS'es, even with different carriage returns, should in fact return true. We should obviously see how different language do that. What do you think?

Eloitor:
from: https://rosettacode.org/wiki/100_doors

> Opening only those doors is an   [optimization](https://rosettacode.org/wiki/Rosetta_Code:Optimization)   that may also be expressed; however, as should be obvious, this defeats the intent of comparing implementations across programming languages.

I'm not against this PR, I just noticed this comment from rosettacode.org :)

PR #1711, opened 2024-09-17
[Logic\all?] Make it work with empty sets

### LINKED ISSUE #1691: [Logic\all?] should return `true` for empty blocks
**Describe the bug**

Right now, this returns *false*. But as @Eloitor rightly pointed out, this is not correct: https://en.wikipedia.org/wiki/Vacuous_truth#In_computer_programming

**See also:** https://discord.com/channels/765519132186640445/829324913097048065/1285292083640402125

drkameleon:
Fixes #1691 

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #1736, opened 2024-09-25
Re-enable `--profiler` mode

drkameleon:
This is just a quick draft... It doesn't really fix anything. Just makes sure we get the `--profiler profile` mode to compile correctly (albeit by disabling e.g. "proc" (which I don't remember what it refers to lol) profiling)


drkameleon:
Ready to merge! 🚀

PR #1753, opened 2024-10-01
[Net\serve] Add support for request handler

### LINKED ISSUE #1752: [Net\serve] add support for Function param / request handler
Right now, `serve` works as some sort of opinionated Sinatra, written internally in Arturo.
This isn't that bad, but it's probably too restricting, being an stdlib function.

What we could do is:

- add support for Function params
- when the user passes a `$[req][...]` parameter, this function will receive a dictionary with all the request details
- and it will expect a request answer (= dictionary) as a response

This way we will allow for even more fine-tuned handling of the whole server.

Also, this would allow whatever Sinatra- or non-Sinatra-like alternative to be developed in the form e.g. of external packages.

As usual: ease of use, yes; but also, flexibility. 
(Arturo should not be about showing us the ONE way to do things, but instead enabling us different paths to... the solution 😉 )

drkameleon:
# Description

This PR will make it possible to do sth like:

```
serve $[req][
   ; do something with the request
   ; and return a response
]
```

Fixes #1752 

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #1758, opened 2024-10-09
Fix custom types example

Eloitor:
# Description

Fix custom types example.

I'm not sure about the comparator, since deleting that line doesn't change anything.

## Type of change

- [ ] Code cleanup

### DIFF
--- a/docs/website/pages/documentation/in a nutshell.art
+++ b/docs/website/pages/documentation/in a nutshell.art
@@ -376,29 +376,26 @@ g: function [x][
 ;---------------------------------
 
 ; defining a custom type
-define :person [                            ; define a new custom type "Person"
-    name                                    ; with fields: name, surname, age
-    surname
-    age 
-][ 
-    ; with custom post-construction initializer
-    init: [
-        this\name: capitalize this\name
+define :person [
+    ; define a new custom type "Person"
+    ; with fields: name, surname, age
+    init: method [name surname age][
+        \name: capitalize name
+        \surname: surname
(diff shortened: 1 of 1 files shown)


drkameleon:
Good idea - we should start dealing with pending documentation issues asap!

Re: comparator.

Right now, comparing object of a given user-defined type is meant to happen either implicitly or using a provided `compare:` magic method.

Now, that magic method, we may define it ourselves (e.g. `compare: method [that][ compare \x that\x ]`, etc), or if it's just about one specific field, we could use the `sortable` helper, like: 
```
compare: sortable 'age
```

And it most definitely influence the output. 🤔 
Have you tried sorting by... a different field?

Eloitor:
What does it do if no sortable method is defined?

drkameleon:
> What does it do if no sortable method is defined?

The answer to the mystery:

https://github.com/arturo-lang/arturo/blob/master/src/vm/values/comparison.nim#L348-L356

So, if you did:

```
customObjectA < customObjectB
```

it would:

- first look if there is a `compare` magic method defined
- then look for `less?`
- and if none of that is defined, it would return `false` (however, `customObjectB < customObjectA` would *also* return false)

drkameleon:
Another example:

Screenshot 2024-10-09 at 14 22 54

Eloitor:
Oh... `# should throw!` ... I agree...

drkameleon (approved):

drkameleon:
Ready to merge! 🚀

PR #1773, opened 2024-10-18
Sanitize `?`s in words/labels

drkameleon:
# Description

Question marks, by convention or not, are meant to add some variant of an existing symbol; or simply denote a predicate (= function that returns a Logical value).

That being said, the only place where this would make sense would be at the very end of a word. (e.g. `something?`). Anything else should be disallowed. And, of course, we shouldn't allow more than one `?` within a word - yes, believe it or not this is absolutely a Word right now: `something?anotherone??`.

Let's start untangling this mess.

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #1774, opened 2024-10-18
Sanitize `_` in words/labels

drkameleon:
# Description

**Also see:** https://github.com/arturo-lang/arturo/pull/1773 (similar-ish)

## Type of change

- [ ] Code cleanup


Eloitor:
Could this be a step in the direction to get negative numbers like `_1`, `_2` or `_3`?

drkameleon:
> Could this be a step in the direction to get negative numbers like `_1`, `_2` or `_3`?

The logic is a bit different... and more for reasons of consistency: simply put, `_` is not a "word". Now, considering `_1` as "not a word" would be tempting too, but given that these are totally valid identifiers e.g. in JSON, this would start degrading the compatibility between what Arturo support and different (and common) data formats. And in face of the long-await (still proof of concept) [Arturo-as-Data](https://github.com/arturo-lang/arturo/pull/1763), we definitely don't want to completely spoil it. (And one could say why not `_` - which is also a valid JS identifier? Because, I guess we'd have to set *some* limit - even though it might seem arbitrary - and exclude things that go against a very core logic in Arturo: if it looks like a symbol, then it's most definitely not a "word" lol)

**A couple of sidenotes:**

- For negative numbers, as a literal, I'm still leaning in favor of your own unary-minus idea... 
- The reason behind this PR - apart of a general effort of cleaning up ambigously parsable values - is this: https://discord.com/channels/765519132186640445/829324913097048065/1277266472120619171 (it's not going to be exactly like this, but our `case` supporting pattern matching could be more feasible and visually-ok than it initially looked 😉

PR #1786, opened 2024-11-09
[Collections] Refactoring Collections

RickBarretto:
Just refactoring some code


drkameleon:
Very interesting approach. And it will help us be able to see more clearly what is going on...

The code is far more readable. (= the side effects of adding and adding things, and the code ending up a patchwork lol)

Can we also make sure there is no performance difference? (I would think that there should be a boost, but let's do some tests before we continue) Can you run it through Peregrino - or anything else for that matter - to see how we fare?

RickBarretto:
@drkameleon It seems that everything is right here.
I don't want to make this PR longer than it should be. I'll open other and continue.
I don't want to unsync that much with our main branch.

RickBarretto:
Please, have a look on the unit-tests and refactoring and let me know if you like it. :wink:

drkameleon:
> Please, have a look on the unit-tests and refactoring and let me know if you like it. 😉

@RickBarretto 
That's some work here! 😉 

Obviously, the unit-tests part... there is nothing to comment on here - the more... the better! :) 🚀 

Re: the refactoring, I'll have to *really* study it, since we are talking about *very* core stdlib functions, so... one mistake and the whole language suffers.

At first sight, I'm not sure I would add functions (`proc`) inside implementations - in this case, they would normally in core VM modules, or as helpers... but we can see about this later.

**My main points are:**

- have you benchmarked the changes? (I mean.. the new ones have to be at least as good - or bad lol - as they were, performance-wise)
- also, I would totally look into the generated C code - whenever I see too much playing around with `ValueArray`'s etc, I would ask: "are you sure there hasn't been an unnecessary copy anywhere? how sure are you?" lol - and that's not even a joke: I've actually had to look into the generated C source 489324 times to make sure things are as I want them to be... and even then we'll miss things. But in general, working with `seq`s requires caution!

drkameleon (commented):
To make it official: 😛 
https://github.com/arturo-lang/arturo/pull/1786#issuecomment-2581840604

drkameleon:
Basically, what I would suggest is:
- take one, isolated function that has been changed (e.g. `array`)
- write something as excruciating thorough and detailed as possible, using this function
- benchmark the script with two versions: (A) a master branch binary, (B) the new binary
- do that at least on your own system and make sure the results are consistent

Now, since this testing should actually be done on all 3 systems (and not whatever OS we may happen to be using), this would make me think that:
- preferrably, we should run the exact same thing on: macOS + Linux
- "how?" - we could fine-tune [Peregrino](https://peregrino.pkgr.art/) and use it inside a CI action? (Would you be up for ... the challenge? 😉 )

RickBarretto:
I'll have a look on how to run this on CI.
I can't test it in multiple OSes, and also testing one by one would be a "longer than it should" be task...
I'll open another PR so, adding benchmarks. I'll look for some project that use this approach.

drkameleon:
> I'll have a look on how to run this on CI. I can't test it in multiple OSes, and also testing one by one would be a "longer than it should" be task... I'll open another PR so, adding benchmarks. I'll look for some project that use this approach.

Personally, I think this way:

- Get a more-or-less official way to benchmark X
- Benchmark X locally (as a quick check to make sure we are not too far off)
- Then find a CI-based way to check the exact same way, cross-platform

P.S. I think sth really interesting could come out of this... 😉

drkameleon:
> * Then find a CI-based way to check the exact same way, cross-platform

And to elaborate on my own idea:

I don't mean that we should have a set of benchmarks running all the time (that we could have, but for measuring e.g. how "Arturo works as a whole"). What I mean is that we should have a trigger-able CI action, let's say, that... when something is set, it would run the appropriate tests... 🤔 

Give me some time and I'll do my best to - perhaps - come up with some draft idea!

RickBarretto:
I still need to reimplement inplace behavior for `couple` and re-implement read from file for `array`. :wink:

PR #1788, opened 2024-11-09
Add new `discard` stdlib function

### LINKED ISSUE #1780: Core: add new `discard` function
**Example:**

```
discard 2
```

it would work pretty much like:

```
2
unstack.discard 1
```

(which doesn't really make sense)

----

Also, another use case (from the Validator package):

```
return not? throws? [
    _: to :date .format:fmt str
]
```
 (we just want to check if it throws, but not just dump the value onto the stack - in that case, we could have used `discard` 😉 )

drkameleon:
# Description

Fixes #1780 

## Type of change


drkameleon:
Ready to merge! 🚀

PR #1731, opened 2024-09-24
Amplify backend actions for webviews

drkameleon:
# Description

Right now, one of the different "access points" we have, so that our webviews communicate with the Arturo are *backend actions*. That is: named, function-like calls, available exclusively in this context (Webview <-> Arturo message passing) and are normally meant to offer system- or UI-specific functionality (e.g. "maximize the main window" ➡️ `window.maximize`).

Here, we'll attempt adding a few more of these actions and, thus, amplify a bit the spectrum of what can be done/accessed through the webview itself! 😉 

## Type of change

- [x] Code cleanup
- [x] Enhancement (implementation update, or general performance enhancements)
- [x] New feature (non-breaking change which adds functionality)

PR #1821, opened 2024-12-07
Add safe `get`/`set` overrides for custom types

drkameleon:
# Description

Right now, two of the different "magic" methods that we can define for custom types are `get` and `set` - and the... possibilities of that are... extreme! 🔥  (similar to Ruby's e.g. `attribute_name=(value)` overrides)

**For example, we can do that:**

```ruby
define :person [
    get: method [what][
        print ["getting:" what]
    ]

    set: method [what, value][
        print ["setting:" what "to:" value]
    ]
]

p: to :person []!
p\name: "John"

; and it will print:
; setting: name to: john
```

⚠️ Now, the problem is quite obvious(?):

if `p\name` triggers `get` and `p\name:` triggers `set`, how would we get or set a value on `this`? (= from inside the object for example)

The answer, right now, would be: we *can't*.

Because if we did this:

```
   set: method [what, value][
        print ["setting:" what "to:" value]
        this\[what]:  value
    ]
```

Guess what: it would trigger an endless loop.

So... let's fix it!

## Type of change


drkameleon:
**Solution:**

```
define :person [
    get: method [what][
        print ["getting:" what]

        (key? this what)? -> get.safe this what ; if the key exists, return the value
                          -> "DEFAULT"          ; do something else
    ]

    set: method [what, value][
        print ["setting:" what "to:" value]
        set.safe this what value
    ]
]


p: to person []!

p\name: "John"

print p\name 
; John

print p\whatever 
; DEFAULT
```

drkameleon:
**Why all the fuss:**

Because this could pave the way for things related to the [UI transformations I'm working on](https://github.com/arturo-lang/arturo/pull/1731).

For example, if we pack all the backend actions in a `window` object...

We could do `window\setTitle` and set the window's title. Or `window\getTitle` or (`window\title`) to get the current window's title.

But: the most Arturo way would be simply... `window\title` (to get it) and `window\title: "whatever"` to set it...

How can we do it, without adding another custom type internally and 3 million if-else-if-else in the `get`/`set` implementations in case we have a Window pseudo-object?

Well... like this! 😉

drkameleon:
Ready to merge! 🚀

PR #1830, opened 2024-12-13
[Sockets\listen] remove `.blocking` option

drkameleon:
# Description

Basically, the only thing Arturo currently supports (and quite well, actually) is *blocking* connections, hence all of the examples so far were with `listen.blocking`.

Instead of doing it like this, this PR:

- completely removes the `.blocking` option
- internally sets it as the default mode

And if in the future we actually add support for *non-blocking* sockets, we could add an `.async` option for that...

## Type of change

- [x] Code cleanup

### DIFF
--- a/src/library/Sockets.nim
+++ b/src/library/Sockets.nim
@@ -133,7 +133,6 @@ proc defineLibrary*() =
                 "port"  : {Integer}
             },
             attrs       = {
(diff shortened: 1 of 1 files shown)


drkameleon:
Ready to merge! 🚀

PR #1842, opened 2024-12-30
[Core] add new `using` stdlib function

### LINKED ISSUE #1793: Resolve Warning: `imported and not used: 'profiler'`
**See also:** https://discord.com/channels/765519132186640445/829324913097048065/1306573998745981021

### LINKED ISSUE #1824: [Core] Add new `using` function
Right now, we have [`with`](https://arturo-lang.io/master/documentation/library/core/with/) which basically creates a new block by encapsulating a given variable/value (very useful in a "closure" context).

What `using` would do is something similar-ish, but:

- safer
- totally scoped
- more specific
- centered around `this`

Also, `using` as a name is used in very similar contexts in other languages... So...

In short: the idea is to... introduce a forced `this` (and restore the previous one, if there), so that we can use field `\shortcuts`.
As similar idea is Nim's own `with`: https://nim-lang.org/docs/with.html#with.m%2Ctyped%2Cvarargs%5Buntyped%5D

And, yes, it's more than just a simple coincidence... lol (👀 https://github.com/arturo-lang/arturo/pull/1731)

**Example:**

```red
; before
window\size: [300, 400]
window\center
window\title: "New window"

; after
using window [
    \size: [300, 400]
    \center
    \title: "New window"
]
```

drkameleon:
# Description

Fixes #1824 
Fixes #1793 (totally unrelated, but since I noticed it...)

## Type of change


drkameleon:
Ready to merge! 🚀

PR #1848, opened 2025-01-06
[Strings\numeric?] make it work consistently

drkameleon:
# Description

See also: https://discord.com/channels/765519132186640445/829324913097048065/1325722145250086923

## Type of change


RickBarretto:
The reason why this behaves that was is very simple.

```
$> numeric? "."
=> true

$> to :floating "."
=> 0.0

$> numeric? "0.0"
=> true
```

`to :floating "."` is parsed to `0.0`. This is not inconsistent, I would say. But, should this behavior like this?

PR #1867, opened 2025-01-22
Update documentation

RickBarretto:
## Type of change

- [ ] Code cleanup

### DIFF
--- a/src/library/Collections.nim
+++ b/src/library/Collections.nim
@@ -1411,6 +1411,15 @@ proc defineLibrary*() =
         },
         returns     = {Range},
         example     = """
+        ; range of :integers
+        array range 0 5     ; [0 1 2 3 4 5]
+        @0..5               ; [0 1 2 3 4 5]
+        ..........
+        ; range of :chars
+        @'a'..'e'           ; [a b c d e]
+        ..........
+        ; range with steps
+        range.step: 2 1 5   ; [1 3 5]
         """:
             #=======================================================
             var limX: int
--- a/src/library/Collections.nim
+++ b/src/library/Collections.nim
@@ -1501,11 +1501,11 @@ proc defineLibrary*() =
             remove.index: 2 "Ruby" "u"  ; => Rby
             remove.index: 2 "Ruby" "a"  ; => Ruby
             ..........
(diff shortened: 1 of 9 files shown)


drkameleon:
I **love** what you're doing here! 😃 

P.S. Don't bother (too much) with `as` & `from` as... as I have said, I consider them too messy and probably scheduled for cleanup or deletion. But other than that: just go for it! Awesome job! 🚀

RickBarretto:
Ok, I think it's done! :wink:

src/library/Collections.nim
```diff
@@ -1403,6 +1411,15 @@ proc defineLibrary*() =
...
         example     = """
+        ; range of :integers
+        array range 0 5     ; [0 1 2 3 4 5]
+        @0..5               ; [0 1 2 3 4 5]
+        ..........
+        ; range of :chars
+        @'a'..'e'           ; [a b c d e]
+        ..........
+        ; range with steps
+        range.step: 2 1 5   ; [1 3 5]
```
drkameleon, line 1422: Can we make this a bit more generic?

I mean... if I read the example, I would think that range always needs `array` (or `@`) which is precisely the point: ranges are "lazy" arrays, that don't become one until/unless needed.

So, if you don't mind, I'd suggest:
- remove the `@`s
- the "results" could be more verbal and not a value (e.g. `; all integers between 0 and 5, inclusive")
- you could add an extra example *with* `@` only to highlight that *this*, yes, is an array (aka "evaluated range" 😉 )

drkameleon (changes requested):
Wow. OK. That's something!! 

Great job! ❤️ (Should I even mention it?!)

Wow. Keep going, very valuable PR! 😉 

P.S. If you agree, just a tiny fix that I marked above...and I'm merging! 🚀

src/library/Collections.nim
RickBarretto: You're completely right, I'll fix it very soon. :wink:

RickBarretto:
Updated! 😉

drkameleon (approved):
Perfect!!!

Let's get it in! 😉

drkameleon:
Ready to merge! 🚀

PR #1902, opened 2025-02-25
Remove Converters module

drkameleon:
# Description

This module shouldn't even exist as it is. Right now, it barely contains 2 functions - and they could be both considered misnomers (or grouping things that shouldn't be grouped).

Either way, ahead of the upcoming release, it would be better if we just moved these functions ... literally *anywhere* so that - at least - we don't have a module with no purpose whatsoever. (Plus, the module itself - not the functions - isn't properly documented as the rest of the modules)

Close when all of the above are merged:

- [x] #1923
- [x] #1924 
- [x] #1925 
- [x] #1926 
- [x] #1927 

Fixes: #1912 

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

Issue #1913, opened 2025-02-27
[Converters\as] Needs sanitization


drkameleon:
## Top priority

**`.binary`, `.hex`, `.octal`** ➞ ❌ *remove*

Instead of:
```ruby
as.binary 10
as.hex 255
as.octal 8
```

We could simply have:
```ruby
to :string .base:2 10          ; "1010"
to :string .base:16         ; "FF"
to :string .base:8 8           ; "10"
```

(with could use `.binary`, `.hex`, `.octal` too - although this would lead to unnecessary clutter for something that isn't even that common!)

Another idea would be using `.format:` which is already supported.

---

**`.code`, `.pretty`, `.unwrapped`** ➞ ✅ *add new*

Essentially, all three modes have to do representing an Arturo value as *code* - with `.pretty` and `.unwrapped` specifying *what* this code will look like.

So, we could create a new function (e.g. `express`, `codify`, or `serialize` - by order of personal preference lol) and keep `.pretty` & `.unwrapped` as options

Instead of:
```ruby
as.code [a: 2]
```

We could simply have:
```ruby
express [a: 2]
```

---

## For future reference

**`.agnostic`** ➞ ✅ *add new*

Create a function with the same name (= `agnostic`)
With the prerequisite, that we are sure what this thing does.

---

**`.data`** ➞ ✅ *add new*

Create a function `parse` (that could be a generic shortcut for `to :block ...` and be used like a basis for creating our (and proper) `parse` equivalent.

`.data` could be an option, like now, to allow for "Arturo as data". (`parse.data ...`)
With the prerequisite, that we are sure what this thing does.

drkameleon:
> [@drkameleon](https://github.com/drkameleon) what does `agnostic` do?

Thanks for reminding me that I forgot this PR open! lol

Well... I had to look it up myself... basically, it took a block, and auto-converted all non-assigned `:word`s to Literal values. Very weird functionality, that a) I guess I needed at some point for some project, b) it can be easily replicated in Arturo itself, c) I couldn't find a single case of it being used, anywhere... so, I just removed it!

PR #1926, opened 2025-03-06
Add new `agnostic` stdlib function

drkameleon:
# Description

See also: #1913 

## Type of change

- [x] Code cleanup


drkameleon:
This PR will be closed.

Unless we actually need it, as far as I can tell this functionality must have been used *once* max, in some package of mine (and I cannot even spot it right now) - not to mention that implementing this ourselves, in Arturo, is like 2 lines of code.

So... until then, let's just remove it altogether. Sometimes all it takes is simply subtracting ;)

PR #1963, opened 2025-05-05
Fix #1873: fold.seed should return the seed for empty lists

### LINKED ISSUE #1873: [Iterators/fold] `fold.seed` should return the seed for empty lists
**Describe the bug**

> `fold.seed:”test” “” [acc chr] [acc ++ chr]` returns an empty block. I would expect the seed value back when dealing with an empty collection.

```
$> fold.seed: "test" "" [acc chr][acc ++ chr]
; => []
```

**Expected behavior**
```
$> fold.seed: "test" "" [acc chr][acc ++ chr]
; => test
```

**Additional context**
Thanks to @BNAndras on Discord:
https://discord.com/channels/765519132186640445/1328439373774983300/1328439373774983300

deepestsound:
Fixes #1873

### DIFF
--- a/src/library/Iterators.nim
+++ b/src/library/Iterators.nim
@@ -923,7 +923,7 @@ proc defineModule*(moduleName: string) =
                 else:
                     let firstElem {.cursor.} = rang[0]
                     res = case firstElem.kind
-                        of Char    :newString("")
+                        of Char    : newString("")
                         of Floating: newFloating(0.0)
                         else       : I0
                
--- a/src/library/Iterators.nim
+++ b/src/library/Iterators.nim
@@ -306,7 +306,7 @@ template prepareIteration(doesAcceptLiterals=true) {.dirty.} =
 template fetchIterableRange() {.dirty.} =
     var rang = iterable.rng
 
-template fetchIterableItems(doesAcceptLiterals=true, hasSeed: bool, defaultReturn: untyped) {.dirty.} =
+template fetchIterableItems(doesAcceptLiterals=true, defaultReturn: untyped) {.dirty.} =
     var blo = 
         case iterable.kind:
             of Block,Inline:
@@ -323,7 +323,7 @@ template fetchIterableItems(doesAcceptLiterals=true, hasSeed: bool, defaultRetur
             else: # won't ever reach here
                 @[VNULL]
 
-    if blo.len == 0 and not hasSeed: 
+    if blo.len == 0 and (when declared(hasSeed):not hasSeed else: true): 
         when doesAcceptLiterals:
             when astToStr(defaultReturn) != "nil":
                 if unlikely(inPlace): RawInPlaced = defaultReturn
@@ -415,7 +415,7 @@ template doIterate(
             itAct
         itPost
     else:
-        fetchIterableItems(doesAcceptLiterals=itLit, hasSeed=false):
+        fetchIterableItems(doesAcceptLiterals=itLit):
             itDefVal
 
         itPre
@@ -654,7 +654,7 @@ proc defineModule*(moduleName: string) =
                     if unlikely(inPlace): RawInPlaced = newBlock(res)
                     else: push(newBlock(res))
                 else:
-                    fetchIterableItems(doesAcceptLiterals=true, hasSeed=false):
+                    fetchIterableItems(doesAcceptLiterals=true):
                         newBlock()
                     
                     iterateBlock(withCap=false, withInf=false, withCounter=false, rolling=false):
@@ -812,7 +812,7 @@ proc defineModule*(moduleName: string) =
                 if unlikely(inPlace): RawInPlaced = newBlock(res)
                 else: push(newBlock(res))
             else: 
-                fetchIterableItems(doesAcceptLiterals=true, hasSeed=false):
+                fetchIterableItems(doesAcceptLiterals=true):
                     newBlock()
 
                 if onlyLast:
@@ -933,7 +933,7 @@ proc defineModule*(moduleName: string) =
                 if unlikely(inPlace): RawInPlaced = res
                 else: push(res)
             else:
-                fetchIterableItems(doesAcceptLiterals=true, hasSeed=hasSeed):
+                fetchIterableItems(doesAcceptLiterals=true):
                     newBlock()
 
                 var res: Value
@@ -1116,7 +1116,7 @@ proc defineModule*(moduleName: string) =
                 if unlikely(inPlace): RawInPlaced = newBlock(res)
                 else: push(newBlock(res))
             else: 
-                fetchIterableItems(doesAcceptLiterals=true, hasSeed=false):
+                fetchIterableItems(doesAcceptLiterals=true):
                     newBlock()
 
                 var res: ValueArray = newSeq[Value](blo.len)
@@ -1358,7 +1358,7 @@ proc defineModule*(moduleName: string) =
                     if unlikely(inPlace): RawInPlaced = newBlock(res)
                     else: push(newBlock(res))
             else: 
-                fetchIterableItems(doesAcceptLiterals=true, hasSeed=false):
+                fetchIterableItems(doesAcceptLiterals=true):
                     newBlock()
 
                 if onlyLast:
--- a/src/library/Iterators.nim
+++ b/src/library/Iterators.nim
@@ -323,7 +323,7 @@ template fetchIterableItems(doesAcceptLiterals=true, defaultReturn: untyped) {.d
             else: # won't ever reach here
                 @[VNULL]
 
-    if blo.len == 0 and (when declared(hasSeed):not hasSeed else: true): 
+    if blo.len == 0 and (when declared(hasSeed): not hasSeed else: true): 
         when doesAcceptLiterals:
             when astToStr(defaultReturn) != "nil":
                 if unlikely(inPlace): RawInPlaced = defaultReturn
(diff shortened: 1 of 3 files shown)


deepestsound:
TESTS:

print equal? :block type fold "" [x y][x ++ y]
print equal? "test" fold "test" [x y][x ++ y]
print equal? "t" fold "t" [x y][x ++ y]
print equal? "hi" fold.seed:"hi" "" [x y][x ++ y]
print equal? "hiyou" fold.seed:"hi" "you" [x y][x ++ y]
print equal? "hit"fold.seed:"hi" "t" [x y][x ++ y]
print equal? "" fold.seed:"" ""  [x y][x ++ y]
print equal? "you" fold.seed:"" "you" [x y][x ++ y]
print equal? "t" fold.seed:"" "t" [x y][x ++ y]
print equal? 's' fold.seed:'s' "" [x y][x ++ y]
print equal? "syou" fold.seed:'s' "you" [x y][x ++ y]
print equal? 1 fold 1..1 [x y][x + y]
print equal? 55 fold 0..10 [x y][x + y]
print equal? 155 fold.seed:100 0..10 [x y][x + y]
print equal? "abc" fold 'a'..'c' [x y][x ++ y]
print equal? "hello abc" fold.seed:"hello " 'a'..'c' [x y][x ++ y]
print equal? "abc" fold.seed:"" 'a'..'c' [x y][x ++ y]

drkameleon (changes requested):
Good job!

I have just one problem: generally, the code in Iterators is ultra-optimized for performance (hence this super-weird way of going about it, and the templates, etc).

I understand the logic behind `hasSeed`.

The problem is: 
- `fetchIterableItems` is used by *all* functions in the module
- but `hasSeed` (and the related tests) are in fact needed by only one function (= `fold`)

My point is:
- we introduce extra, runtime, complexity (e.g. `if blo.len == 0 and not hasSeed:`) for many function that don't really need it (and, in all probability, are far more commonly used, compared to `fold`).

Can you find a way to:
- either make `hasSeed` compile-time (generally: if you are using `if`s it's runtime; `when` is compile-time)?
- or isolate the changes in `fold`'s implementation? (= without affecting other helper functions/templates)

Your changes are very valid. I'd just suggest we integrate them in a way that doesn't negatively impact - in any way (not that it's perfect, huh?) things that are already working more-or-less fine. 😉

drkameleon:
> TESTS:
> 
> print equal? :block type fold "" [x y][x ++ y] print equal? "test" fold "test" [x y][x ++ y] print equal? "t" fold "t" [x y][x ++ y] print equal? "hi" fold.seed:"hi" "" [x y][x ++ y] print equal? "hiyou" fold.seed:"hi" "you" [x y][x ++ y] print equal? "hit"fold.seed:"hi" "t" [x y][x ++ y] print equal? "" fold.seed:"" "" [x y][x ++ y] print equal? "you" fold.seed:"" "you" [x y][x ++ y] print equal? "t" fold.seed:"" "t" [x y][x ++ y] print equal? 's' fold.seed:'s' "" [x y][x ++ y] print equal? "syou" fold.seed:'s' "you" [x y][x ++ y] print equal? 1 fold 1..1 [x y][x + y] print equal? 55 fold 0..10 [x y][x + y] print equal? 155 fold.seed:100 0..10 [x y][x + y] print equal? "abc" fold 'a'..'c' [x y][x ++ y] print equal? "hello abc" fold.seed:"hello " 'a'..'c' [x y][x ++ y] print equal? "abc" fold.seed:"" 'a'..'c' [x y][x ++ y]

Btw, you got add this under [tests/unitt/lib](https://github.com/arturo-lang/arturo/tree/master/tests/unitt/lib), but since @RickBarretto hasn't already added a folder for Iterators - it could be the first one there.

P.S. I swear I'm not that familiar with our unit-test system either, but... getting used to it! 😛  lol

drkameleon (approved):
Love the minimalism in the change now! Very nicely done! :) 🚀 

If you don't mind, I'll ping @RickBarretto too for a check (re: the unit-tests), because that part is all his work -- although with all honesty, it looks good and if the tests are passing, there should be nothing else... to fix. 😉

tests/unitt2/lib/iterators/fold.test.art
```diff
@@ -0,0 +1,69 @@
+import.version:2.0.1 {unitt}!
```
RickBarretto, line 1: Currently, we are not using Unitt v2 on our codebase, I've tried to include it, but in the end I thought this was not prepared for a big codebase like this.
So, I would try to implement using Unitt v1, as the other tests. I'll be updating everything when the v3 came out.

I believe the fact of v2 be running using only one interpreter was not a good choice. I mean, for few modules, it's ok. But for larger one, this is probably not suitable, such as our case.

RickBarretto: About the tests itself. Very good work, well written and readable. I can pretty much understand what is the desirable behavior here. :wink:

RickBarretto: @deepestsound If you need help to implement it for v1 or want me to do this, let me know. I'm open to help you. 🙂

tests/unitt2/lib/iterators/fold.test.art
```diff
@@ -0,0 +1,69 @@
...
+
+describe "Allowed Operations" [
+
+    it "fold with no seed and empty string" [
+        expects -> equal? :block type fold "" [x y][x ++ y]
```
RickBarretto, line 6: Just an Arturo tip, you may use `block? <expression>` instead of `equal? :block type <expression>`.

src/library/Iterators.nim
```diff
@@ -908,52 +908,48 @@ proc defineModule*(moduleName: string) =
...
-                    seed = aSeed
-
-                res = seed
-                
+                if hasSeed:
+                    res = aSeed
+                else:
+                    let firstElem {.cursor.} = rang[0]
+                    res = case firstElem.kind
+                        of Char    : newString("")
```
RickBarretto, line 926: Should this be `Char, String` just like the second one, no?
I may be missing something here. @drkameleon knows better about the `fold` function working.

RickBarretto: Also, good idea to use `case/of` instead of `if/else`. 😉 
Great job! 🚀

RickBarretto (changes requested):
I loved your work. But unitt2 should not be used on Arturo codebase. I'll be updating everything to v3 when I launch it.

In general, very good job, well written and understandable, including tests which are fundamental to our codebase.

If CI is working well, @drkameleon approves and tests being downdated to v1, to me, it's ready to merge! 🙂

src/library/Iterators.nim
drkameleon: > Should this be `Char, String` just like the second one, no? I may be missing something here. @drkameleon knows better about the `fold` function working.

I think it's ok - given that `res` is the "accumulator". in that case, think of it this way: if we start with a Char and start adding this value to a "store", this store is actually a String. So, I believe there is no issue. But good to bring it up either way! 😉 Better safe than sorry!

deepestsound: This tests when the iterable is of `range` type, it can never be `string` :)

tests/unitt2/lib/iterators/fold.test.art
deepestsound: Good call

tests/unitt2/lib/iterators/fold.test.art
deepestsound: I do need help unitt v 1.1.2 is bugged and never works for me, even when copying existing tests. That's why I used v2, it's the only one that I got to work

src/library/Iterators.nim
drkameleon: > This tests when the iterable is of `range` type, it can never be `string` :)

Geez, yes! Spot on! :) (The side-effects of not having re-read the whole block of code haha)

tests/unitt2/lib/iterators/fold.test.art
RickBarretto: Ohhh, good catch. Can you tell me what don't work with unitt v1? I'll try to run it later on v1 and v2 and see what happens. Not right now, but when I come home today :wink:.

If we can't make it work for v1, so, v2 is approved for this PR. :wink:

src/library/Iterators.nim
RickBarretto: > This tests when the iterable is of `range` type, it can never be `string` :) 

Good catch!

tests/unitt2/lib/iterators/fold.test.art
deepestsound: Check [these](https://github.com/arturo-lang/arturo/actions/runs/14913837884/job/41894485627) unitt tests logs. It's like this on my machine too.

RickBarretto: Ohh, this is because unitt v1 was created when Arturo had `as` function. Internally, we use `as.code` a lot to show the assertion code in the terminal.

Well, all set up, so. PR approved.

This is not a bug of unitt v1, but an incompatibility with the newer version of Arturo.
Which reminds me to finish v3 very soon to fix all this. :wink:

RickBarretto:
@drkameleon approved! 😉 
Ready to merge?

tests/unitt2/lib/iterators/fold.test.art
deepestsound: Oh, that makes sense! Are you going to temporarily change Ci.yaml to include those tests running on unitt v2?

deepestsound: When v3 is ready just @ me and I'll migrate all tests to v3

drkameleon:
> @drkameleon @deepestsound, approved! 😉 Ready to merge?

Awesome job! 😃

drkameleon (approved):
Ready to merge! 🚀

PR #1964, opened 2025-05-06
Fix #1875 symbol literal not parsed correctly

### LINKED ISSUE #1875: [VM\Parse] symbol literal not parsed correctly
## Description

We know that we have predicates in arturo, and this is treated as any other identifier: `even?`, `is?`, `ok?` are all considered words and will works as any other variable. We also know that adding the `'` prefix, this would return a `:literal`, so: `'even?`, `'is?`, `'ok?` are all `:literal`s which permits lazy-ly (in the sense of: deferred, postponed) evaluation.

The same would be applied to `:symbol`s when prefixed by `'` (that are `:symbolLiteral`), so this would also be lazy-ly evaluated.

But this is not being parsed properly

## Demonstration

**Question-Mark**

Bringing this edge-case because not only this is wrongly parsed, this is considered other type: `:literal`.

```
$> inspect [?]
[ :block
        ? :symbol
]
$ inspect '?
? :literal
```

I would expect:

```
$ inspect '?
? :symbolLiteral
```

**Double Question-Mark**

This edge-case considers `'??` as being two different values, since `'?` is parsed as a `:literal`, then the next `?` is parsed as a `:symbol`.

```
$> inspect [??]
[ :block
        ?? :symbol
]

$> inspect '??

══╡ Value Error ╞═════════════════════════════════════════════════════════════════════════════════════════════ <repl> ══

  Not enough parameters

  Cannot perform:
      switch

  Required:
      3

=> ?

$> inspect ['??]
[ :block
        ? :literal
        ? :symbol
]
```

**Exclamation-Mark**

The same applies to `'!`:

```
$> inspect '!
! :symbolliteral

$> inspect ['!]
[ :block
        ! :symbolliteral
]

$> inspect '!=

══╡ Value Error ╞═════════════════════════════════════════════════════════════════════════════════════════════ <repl> ══
  Not enough parameters

  Cannot perform:
      equal?

  Required:
      2

=> !

$> inspect ['!=]
[ :block
        ! :symbolliteral
        = :symbol
]
```

## Observation

It seems that the issue is due to the first token being considered a `:literal` and other multi-char tokens are being parsed correctly:

```
$> inspect ['<]
[ :block
        < :symbolliteral
]

$> inspect '<
< :symbolliteral

$> inspect ['<=]
[ :block
        <= :symbolliteral
]

$> inspect '<=
<= :symbolliteral
```

**Desktop (please complete the following information):**
```
arturo v/0.9.84-alpha b/3334 (amd64/windows)
```

deepestsound:
Fixes #1875

### DIFF
--- a/src/vm/parse.nim
+++ b/src/vm/parse.nim
@@ -509,10 +509,12 @@ template parseIdentifier(p: var Parser, alsoAddCurrent: bool) =
     when alsoAddCurrent:
         add(p.value, p.buf[pos])
     inc(pos)
+    let initialPos = pos
     while p.buf[pos] in PermittedIdentifiers_In:
         add(p.value, p.buf[pos])
         inc(pos)
-    if p.buf[pos] == QuestionMark:
+    # '?' by itself shouldn't be a valid identifier   
+    if p.buf[pos] == QuestionMark and pos != initialPos:
         add(p.value, QuestionMark)
         inc(pos)
     p.bufpos = pos


drkameleon (approved):
That's a very nice one! Accurate and surgical! 😉

And for confirmation:
Screenshot 2025-05-06 at 12 37 56

Well done! :)

drkameleon:
Ready to merge! 🚀

RickBarretto:
@deepestsound lovely! 🚀

PR #1985, opened 2025-06-23
[Iterators] Sanize function signatures & documentation

### LINKED ISSUE #1932: [Iterators] Sanitize & verify function descriptions
**Describe the bug**

There may be more than a few function descriptions that may be slightly off.

One would be `map` where it actually applies the result of the given block - and, while it's initially referred to as an "action" (which is correct), we then call the argument `condition` (which must be a leftover from copying e.g. `select` or sth along the lines...)

**See also:** https://discord.com/channels/765519132186640445/829324913097048065/1355521112330735769

drkameleon:
# Description

Fixes #1932 

## Type of change

- [x] Code cleanup

### DIFF
--- a/src/library/Iterators.nim
+++ b/src/library/Iterators.nim
@@ -716,7 +716,7 @@ proc defineModule*(moduleName: string) =
         alias       = unaliased,
         op          = opNop,
         rule        = PrefixPrecedence,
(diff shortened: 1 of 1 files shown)


drkameleon:
Ready to merge! 🚀

PR #2009, opened 2025-11-01
[Paths] add new `absolute` stdlib function

### LINKED ISSUE #1997: [Paths\relative] returns input unchanged in script mode (works in REPL)
**Describe the bug**
The [relative ](https://arturo-lang.io/documentation/library/paths/relative/)function works correctly in interactive/REPL mode (it returns a path joined with the current script location), but in script mode it does not resolve the path — it returns the original argument unchanged. This occurs across Windows, Linux (WSL) and Android, and on multiple versions (including v9.83).

**Steps to reproduce**

Create a script with: `print relative "test.txt"`
Run the script (not in interactive REPL).
Actual behavior Output is: test.txt

Expected behavior Output should be the resolved path based on the script location, for example: /Users/admin/Desktop/test.txt

**Environment**

OS: Windows, Linux (WSL), Android (all tested)
Arturo versions: multiple, the newest master branch,and v0.9.83 (issue reproduced across versions)

**Additional context**

Observed consistently on different platforms and versions.
Works as expected in interactive/REPL mode but not when running scripts.

In the [Source ](https://github.com/arturo-lang/arturo/blob/v0.9.83/src/library/Paths.nim#L348)`push(newString(joinPath(env.currentPath(), x.s)))`

Suspected cause It looks like env.currentPath() is not correctly populated when running in script mode, so joinPath(env.currentPath(), ...) resolves to nothing and the function ends up returning the original string.

### LINKED ISSUE #1999: Paths: introduce new `absolute` stdlib function?
The idea/functionality could be something similar to the one described here:
https://github.com/arturo-lang/arturo/issues/1997#issuecomment-3431513125

drkameleon:
# Description

The idea is very closely following this paradigm: https://github.com/arturo-lang/arturo/issues/1997#issuecomment-3431513125

Fixes #1997 
Fixes #1999 

## Type of change


drkameleon:
⚠️ **TODO:** The REPL messes up the frame paths (while it seems to be working fine from a standalone script):

Screenshot 2025-11-01 at 14 48 29

Also mentioned [here](https://github.com/arturo-lang/arturo/issues/1997#issuecomment-3430643662):

> Now, the question would be the opposite: why doesn't the REPL return a relative path?
Well, this is tricky. Because, what is the script being run? (actually it's an embedded script inside Arturo itself - so, should we use Arturo's own binary path as the reference?) it's quite tricky... Sure thing is, I do understand why it looks inconsistent (and you definitely gave me a good reason to re-visit this part). My only point is: the inconsistency actually comes from the way the REPL handles it, not the other way around! 😉

PR #2016, opened 2025-11-05
Display error messages on STDERR

Yewzir:
Nim includes `stderr.writeLine` for that specific purpose.

### DIFF
--- a/src/vm/errors.nim
+++ b/src/vm/errors.nim
@@ -152,6 +152,12 @@ func `~~`*(s: string, inputs: seq[string]): string =
 
 # Error messages
 
+proc printError(s: string) =
+    when not defined(WEB):
+        stderr.writeLine s
+    else:
+        echo s
+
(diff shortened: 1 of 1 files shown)


drkameleon (changes requested):
@Yewzir Good morning!

And thanks for the PR! You are right about `stderr` - let's go for it!

I'm seeing errors with the Web build mostly. Could you look into it?
Since we mostly use `when defined(WEB)` to verify whether we are in a Web/JS build, perhaps a template could be of use, e.g.

```nim
proc errEcho(s: string) =
    when not defined(WEB):
        stderr.writeLine s
    else:
        echo s
```

...and then we'd use that (`errEcho`) instead throughout the file. 

What do you think? 😄

drkameleon:
> Fun fact: this is the 666th closed pull request in the repository. 😈

Yes indeed! 🤣

PR #2055, opened 2025-11-28
Completely remove "safe" builds

drkameleon:
# Description

If everything goes well and we end up transitioning to FreeBSD, we could probably set up the whole thing without needing to compile "limited"/pseudo-sandboxed binaries ourselves. (Which, on top of it, don't appear to be working *that* well anyway...)

## Type of change

- [x] Code cleanup


drkameleon:
This should be considered mergeable -- if and only if the FreeBSD migration succeeds, not before!

drkameleon:
Ready to merge! 🚀

PR #2122, opened 2025-12-24
[Strings\join] add `join.lines` attribute

### LINKED ISSUE #2025: [Strings\join] add `.lines` option
```red
lines: ["line1" "line2" "line3"]
join.lines lines              ; "line1\nline2\nline3"
```

This would work like the conceptual opposite of the existing [`split.lines`](https://arturo-lang.io/master/documentation/library/collections/split/)

RickBarretto:
# Description

Allow user to join lines. See the issue below:

- Fixes #2025

## At A Glance

```art
$> express join.lines ["# Recipe", "", "1. Apple", "2. Banana"]
=> {
        # Recipe
        
        1. Apple
        2. Banana
}
```

## Type of change

- [ ] Code cleanup

### DIFF
--- a/src/library/Strings.nim
+++ b/src/library/Strings.nim
@@ -305,6 +305,8 @@ proc defineModule*(moduleName: string) =
             join.words ["This" "is" "a" "sentence."]
             ; => "This is a sentence."
             ..........
+            ; This example uses the universal new line character
+            ; If you need to use carriage return for some reason, use join.with: "\\r\\n" instead.
             $> join.lines ["# Recipe", "", "1. Apple", "2. Banana"]
             => # Recipe
 


drkameleon (changes requested):
I think you've forgotten to add `lines` as an extra attribute. 😉  (unless I'm missing sth myself with the diffs).

Also, just food for thought, not even sure about it myself: could there be any caveat regarding just using `\n`? It sure is the normal/instictive thing I would think of but... I mean there is some mess regarding what is a "newline" in general, including different definitions by OS, e.g [`\n` vs `\r` vs `\r\n`](https://nim-lang.org/docs/strutils.html#Newlines) (obviously, here it's not about recognizing newlines, it's about injecting them - I just don't know if there's anything else recommended apart from `\n`)

RickBarretto:
@drkameleon I have made some research about this `\r`. I know about its existence, and the only time I have used it was when I did [this PBL](https://github.com/RickBarretto/Eleventh) for my university, that is an HTTP protocol parser. Since HTTP protocol requires it.

But even knowing that Windows uses it due to DOS backwards compatibility, this is a bit strange to me as a programmer, on Windows, because I almost don't care about this Carriage Return. So, I searched and here is the reason:

Python uses universal new line `'\n'` normally and only when you write it to IO, this translates to `\r\n`. Which is a decision that does makes sense, since this would avoid to have parsing issues using regex, for instance, or some inconsistency... See: https://docs.python.org/3/library/functions.html#open

If you read the official Microsoft's documentation example, it uses `\n` internally: https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=msvc-170#example-1

This convention seems to be followed by Nim: https://nim-lang.org/docs/strutils.html

@drkameleon what do you think?

drkameleon:
> @drkameleon I have made some research about this `\r`. I know about its existence, and the only time I have used it was when I did [this PBL](https://github.com/RickBarretto/Eleventh) for my university, that is an HTTP protocol parser. Since HTTP protocol requires it.
> 
> But even knowing that Windows uses it due to DOS backwards compatibility, this is a bit strange to me as a programmer, on Windows, because I almost don't care about this Carriage Return. So, I searched and here is the reason:
> 
> Python uses universal new line `'\n'` normally and only when you write it to IO, this translates to `\r\n`. Which is a decision that does makes sense, since this would avoid to have parsing issues using regex, for instance, or some inconsistency... See: https://docs.python.org/3/library/functions.html#open
> 
> If you read the official Microsoft's documentation example, it uses `\n` internally: https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=msvc-170#example-1
> 
> This convention seems to be followed by Nim: https://nim-lang.org/docs/strutils.html
> 
> @drkameleon what do you think?

well from what I'm seeing the conclusion would be that it's perfectly fine to use `\n` 😉 

(great research btw!! 🚀 )

drkameleon (commented):
Interesting idea!!

drkameleon (approved):
All perfect!!

Awesome job - as usual! 😄

drkameleon:
Ready to merge! 🚀

PR #2128, opened 2025-12-29
Add "build bump" workflow

drkameleon:
# Description

That was previously included in our "Update" workflow, but since most of this logic has move to @arturo-lang/website's deployment workflow, we'll have to add a minimalist one here that does precisely that.

## Type of change


drkameleon:
Ready to merge! 🚀

PR #2144, opened 2026-01-03
[Core] enable a limited `import` even in MINI builds

drkameleon:
# Description

Right now, the package manager and `import` are completely absent in MINI builds. That's mainly because of the lack of SSL-related features and the impossibility of accessing our package server.

However, `import` also supports local "packages" - and that could be a workaround if we wanted to e.g. "externally" download a package and use it via Arturo (even in the MINI version).

## Type of change


drkameleon:
Ready to merge! 🚀

PR #2200, opened 2026-03-03
Introduce Nim-level unit-tests

drkameleon:
# Description

Our tests are - relatively - fine, but having Nim-level, per-file unit-tests wouldn't be bad at all - especially, for critical, lower-level (see: helpers/) implementations. This is even more important since we'll be able to catch core issues, circumventing Arturo (and potential unrelated bugs) altogether.

A rather functional - and tested - solution we could use could be: http://nim-lang.org/docs/unittest.html (= No overcomplicating our lives, can live in the exact same file as the implementation and can be essentially guarded behind a `when isMainModule:` clause, so ... win-win!)

Another idea - as suggested in the docs too - could be simple `assert` statements...

## Type of change


RickBarretto:
That is actually great!

PR #2205, opened 2026-03-07
Unify & clean up main config

drkameleon:
# Description

Right now, apart from the general mess and various things that are not relevant anymore (related to the build script too ofc, that will be rewritten in due time!), our config is too spread out. The result is:

- builds are working fine (because they combine things from the build script + the included config.nims)
- if you e.g. try to [run any dependency-heavy Nim-based unit-tests ](https://github.com/arturo-lang/arturo/pull/2200)(e.g. windows/gmp-related), then we may have issues since it pulls the main config.nims but not other - relevant - parts from (some place inside) the build script.

The goal of this PR is simple:

```
        nim 
        --verbosity:0 
        --warning:all:off 
        --hint:all:off 
        --path:src 
        r 
        src/helpers/bignums.nim
```

should work without issues, on *all* OSes, as-is.       

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #2221, opened 2026-04-18
When evaluating code directly, we should show all errors properly

### LINKED ISSUE #2018: Unhandled IOError in error reporting when evaluating code directly
**Describe the bug**

**See also:** https://github.com/arturo-lang/arturo/issues/1802#issuecomment-3491897650

**To Reproduce**

```
arturo -e "print 1+a"     
```

**Screenshots**

Image

drkameleon:
# Description

The main reason of the crash is that in direct (`-e`) evaluation, there is no code file and thus we shouldn't show any code preview at all.

Fixes #2018 

## Type of change

### DIFF
--- a/src/vm/errors.nim
+++ b/src/vm/errors.nim
@@ -42,8 +42,8 @@
 #=======================================
 
 when not defined(WEB):
-    import re, terminal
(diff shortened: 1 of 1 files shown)


drkameleon:
Ready to merge! 🚀

PR #2225, opened 2026-04-27
Improve function scope performance

drkameleon:
# Description

This PR aims to improve non-inline function call speed. The main logic: 

- instead of cloning the entire Syms table on every call (and replacing it back on exit), we now push a small per-call frame that records (= lazily, only on first write) the previous value of any symbol the function actually touches

- on exit, we simple "replay" that frame in reverse, deleting symbols that didn't exist before the call and restoring the rest, while letting anything in `.exports:` leak through. 

- frames come from a small pool so we don't allocate on every call

The main effect of all this is to significantly decrease the cost per function call and - per my calculations - have a speed gain of ~30%. (Here, on my '18 MacBook at least, the benchmarks when running the main test suite (@ tools/tester.art) have consistently shown a drop from ~230s to 160s) 🚀 

## Type of change

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,8 +110,8 @@ jobs:
             - if: env.arturoCanRun == 'true' && steps.check_skip.outputs.skip != 'true'
               name: Run tests
               run: |
-                cd $GITHUB_WORKSPACE/arturo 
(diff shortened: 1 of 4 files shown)


drkameleon:
Ready to merge! 🚀

PR #2226, opened 2026-04-28
Improve arithmetic operations

drkameleon:
# Description

Right now, all arithmetic operations related opCode (`opAdd`, `opSub`, etc) go through the main pipeline:

- exec loop
- call to closure
- stdlib implementation (with `generateOperationA`/`generateOperationB` that trigger the corresponding branching & operations)

The idea of this PR is very simple: since most arithmetic operations are between normal (= non-bigint) integer, why not include a "fast path" inside the exec loop itself and only resort to the normal call, if that fails?

Per my calculations, the average performance gain is ~5%, while in integer-heavy operations, this can go up to 10%.

## Type of change


drkameleon:
Ready to merge! 🚀

PR #2232, opened 2026-05-01
[Reflection\inspect] Improve functionality & clean up

drkameleon:
# Description

Right now, `inspect` works just fine. The problem is there is a list of long-standing issues:

- formatting is wobbly to say the least (with too much whitespace everywhere)
- nested structures aren't being printed that well
- it doesn't really take any option (we could allow for just one or two)
- some values aren't printed at all (?)

## Type of change

- [x] Code cleanup


drkameleon:
Ready to merge! 🚀

PR #2242, opened 2026-05-28
Convert `do.async` to eager

drkameleon:
# Description

Basically, until now, `do.async` just created a task and added it to the queue. Without any trigger (e.g. `wait`, or `pause`, etc) it would never run.

With this PR, the goal is to convert the command to "eager" (= not lazy) and run it immediately, also adding a new `.lazy` option that corresponds to the previous functionality.

## Type of change

- [x] Code cleanup
- [x] New feature (non-breaking change which adds functionality


drkameleon:
Ready to merge! 🚀

